PR #24, opened 2022-03-08
Create Deadline class

### LINKED ISSUE #23: Create Deadline class
Before implementing the app logic and interface, we should first have a minimal definition of a deadline to work with.

felko:
This PR defines the `Deadline` class defined by a title, a state (currently only TODO or DONE), a start date and an end date.
The dates are `LocalDate`s which means it doesn't let us define deadlines at a specific hour for example.
It also means that the dates are location dependent (since they don't contain the timezone).
The class is therefore subject to change in the future but gives us a basis to get started working on other features.

Closes #23


LennyDZ:
Why do you need to change minSDK from 24 to 26 ??
Maybe use localDateTime instead of localDate, so we can add a time (hh:mm:ss) to the deadline.
Why do we need a start date for a deadline? It will in TODO state as soon as it is create.
Dont remove file on feature commit (I think you removed ExampleUnitTest.kt) this is not good.

felko:
> Why do you need to change minSDK from 24 to 26 ??

The change is motivated in the commit message which you can read by clicking "..." after the commit. Quoting: "The minimum SDK version was updated from 21 to 26 to enable the use of the `Period.between` static method".

I can revert this change and write `between` manually if you think we should rather support lower SDKs.

> Maybe use localDateTime instead of localDate, so we can add a time (hh:mm:ss) to the deadline.

Yes, ultimately this is what we need for later features, however there is no equivalent of `Period` (which is a date interval) for datetimes. I was thinking that I could introduce a new class combining `Period` and `Duration` to represent datetime intervals in a later PR, but I can do it in this PR if you think that it's preferable.

> is the naming attribute "isDue" good? since if we passed the deadline it will be too late, so maybe name it isExceeded or something more "dramatic" than isDue :)

It's unhealthy to worry about things you can't change, if the deadline has passed, there's nothing you can do anymore.
More seriously, I think the name is alright, I reused the terminology in Moodle assignment pages.

> Dont remove file on feature commit (I think you removed ExampleUnitTest.kt) this is not good.

As it was a sample file from the bootcamp it didn't seem problematic to remove it. I can revert this change as well if you really think it would be better. We have to reformat code to use LF instead of CRLF line separators anyway, we could take care of removing the ExampleUnitTest.kt file at the same time in another PR.

ValentinAebi (commented):
## Deadline class
* I think that classes of the model should be immutable (unless good reason for not doing so, of course). Here it is probably not good to allow mutations of the attributes of Deadline. The risk of doing a mistake is much lower with an immutable class. Of course the user may want to modify a deadline on the app, but in this case we can replace the stored Deadline object instead of mutating it (Kotlin has a built-in `copy` method for data classes, which can be handy in this case).
* It may be better to define `Deadline` as an interface, so that we can consider multiple types of deadlines (or tasks). E.g. storing a start date may not make sense for some kinds of deadlines.
* What if title is the empty string?

## Tests
Using `LocalDate.now()` to obtain the test inputs is problematic: if a bug only occurs when some particular condition on the input time is met, then a failure case of the test may be really difficult to reproduce.

Voxanimus (approved):
On a déjà eu des discussion irl là dessus, le nécessaire me semble là pour le moment.

Matathiasq13824 (approved):
Comme dit @Voxanimus, la classe semble avoir le nécessaire pour la suite

PR #27, opened 2022-03-10
Adding the QR Code feature

macintosh1398:
Adding the first draft of a QR Code reader. A new QR-Code reader activity was created, launched by a button on the main activity, that is able to read and display the result of a QR-Code scan as a "Toast" text.

### DIFF
--- a/app/src/androidTest/java/com/github/multimatum_team/multimatum/QRCodeReaderActivityTest.kt
+++ b/app/src/androidTest/java/com/github/multimatum_team/multimatum/QRCodeReaderActivityTest.kt
@@ -1,0 +1,24 @@
+package com.github.multimatum_team.multimatum
+
+import androidx.test.core.app.ActivityScenario
+import androidx.test.espresso.Espresso
+import androidx.test.espresso.assertion.ViewAssertions
+import androidx.test.espresso.matcher.ViewMatchers
+import androidx.test.rule.GrantPermissionRule
+import org.junit.Rule
+import org.junit.Test
+
+class QRCodeReaderActivityTest {
+    @get:Rule
+    val permissionRule: GrantPermissionRule = GrantPermissionRule.grant(
+        android.Manifest.permission.CAMERA
+    )
+
+    @Test
+    fun should_display_scanner_when_permissionsAreGranted() {
+        ActivityScenario.launch(QRCodeReaderActivity::class.java)
+
+        Espresso.onView(ViewMatchers.withId(R.id.scanner_view))
+            .check(ViewAssertions.matches(ViewMatchers.isDisplayed()))
+    }
+}
--- a/app/src/main/java/com/github/multimatum_team/multimatum/MainActivity.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/MainActivity.kt
@@ -1,10 +1,13 @@
 package com.github.multimatum_team.multimatum
 
 import android.content.Intent
+import android.content.pm.PackageManager
 import androidx.appcompat.app.AppCompatActivity
 import android.os.Bundle
 import android.view.View
 import android.widget.EditText
+import androidx.core.app.ActivityCompat
+import androidx.core.content.ContextCompat
 
 const val EXTRA_NAME = "com.github.multimatum_team.multimatum.main.name"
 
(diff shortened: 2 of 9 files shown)


app/src/androidTest/java/com/github/multimatum_team/multimatum/QRCodeReaderActivityTest.kt
```diff
@@ -0,0 +1,24 @@
...
+import org.junit.Test
+
+class QRCodeReaderActivityTest {
+    @get:Rule
+    val permissionRule: GrantPermissionRule = GrantPermissionRule.grant(
+        android.Manifest.permission.CAMERA
+    )
+
+    @Test
+    fun should_display_scanner_when_permissionsAreGranted() {
```
felko, line 18: You can use this syntax to explain the test scenario in plain english:
```kotlin
fun `Should display scanner when permissions are granted`() {
    ...
}
```

app/src/main/java/com/github/multimatum_team/multimatum/MainActivity.kt
```diff
@@ -1,10 +1,13 @@
 package com.github.multimatum_team.multimatum
 
 import android.content.Intent
+import android.content.pm.PackageManager
 import androidx.appcompat.app.AppCompatActivity
 import android.os.Bundle
 import android.view.View
 import android.widget.EditText
```
felko, line 8: I think this import is not used

app/src/main/java/com/github/multimatum_team/multimatum/QRCodeReaderActivity.kt
```diff
@@ -0,0 +1,55 @@
...
+import android.os.Bundle
+import android.widget.Toast
+import com.budiyev.android.codescanner.AutoFocusMode
+import com.budiyev.android.codescanner.CodeScanner
+import com.budiyev.android.codescanner.CodeScannerView
+import com.budiyev.android.codescanner.DecodeCallback
+import com.budiyev.android.codescanner.ErrorCallback
+import com.budiyev.android.codescanner.ScanMode
+
+class QRCodeReaderActivity : AppCompatActivity() {
```
felko, line 13: Would be nice to have a brief description of the activity

felko (changes requested):

app/src/main/java/com/github/multimatum_team/multimatum/MainActivity.kt
felko: Sorry wrong line, I meant the `import android.widget.EditText` import

app/src/androidTest/java/com/github/multimatum_team/multimatum/QRCodeReaderActivityTest.kt
ValentinAebi: @felko maybe be careful with this "plain english" syntax. I have already had problems with spaces not recognized and replaced by codes in error messages in this syntax, which is unpleasant to read. I would leave it with "_" (but it is a detail).

app/src/main/java/com/github/multimatum_team/multimatum/QRCodeReaderActivity.kt
```diff
@@ -0,0 +1,55 @@
...
+                generateCallbackTest("Camera initialization error: ${it.message}")
+            }
+        }
+
+        // Start the scanner
+        scannerView.setOnClickListener {
+            codeScanner.startPreview()
+        }
+    }
+
```
ValentinAebi, line 51: The name of this method looks a bit strange. Would it not be less confusing to name it e.g. `displayOnToast`?

app/src/main/java/com/github/multimatum_team/multimatum/QRCodeReaderActivity.kt
```diff
@@ -0,0 +1,55 @@
...
+    fun generateCallbackTest(text: String){
+        Toast.makeText(this, text, Toast.LENGTH_LONG).show()
```
ValentinAebi, line 53: The name of this method looks a bit strange. Would it not be less confusing to name it e.g. `displayOnToast`?

app/src/androidTest/java/com/github/multimatum_team/multimatum/QRCodeReaderActivityTest.kt
macintosh1398: For now I will let it like that. Maybe change in the future

felko (approved):
I approve even though the coverage has decreased a lot, we expect that roboelectric will let us write more tests for the camera

macintosh1398:
The QR-Code reader works and can be launched using a button on the main screen. Manual tests were also done to check the ability of the camera to scan QR-Codes. I have improved the coverage from 40% to 87% this week, which is a lot better. For now I don't think I can create more tests but in the future we'll see (more comfortable with UI tests).

PR #28, opened 2022-03-10
Feature/basic settings

### LINKED ISSUE #20: Create activity for settings

ValentinAebi:
I implemented a basic activity for settings. It is able to store settings and retrieve them the next time the app is executed. It currently has a switch for dark mode and another one for notifications, both doing nothing else than storing their values to SharedPreferences (for now, of course).
The tests are basic. They consist in writing some values to the SharedPreferences, modify them from the activity and checking that the buttons are in the correct position and the correct values are stored.
I initially wanted to make dependency injection using a mock SharedPreferences, but after many tries I did not manage to make this work. I pushed my attempts to [a new branch](https://github.com/multimatum-Team/multimatum/tree/feature/basic-settings-mockito-tests), in case someone wants to see whether we can use this kind of tests, but I think that it creates more boilerplate code than it is useful...

Closes #20


app/src/main/java/com/github/multimatum_team/multimatum/MainActivity.kt
```diff
@@ -5,10 +5,12 @@ import androidx.appcompat.app.AppCompatActivity
 import android.os.Bundle
 import android.view.View
 import android.widget.EditText
+import android.widget.ImageButton
```
felko, line 8: This import is not used

felko (changes requested):

felko (approved):

Matathiasq13824 (approved):

PR #36, opened 2022-03-16
Feature/list UI

Matathiasq13824:
Added an Adapter class to show precisely the deadlines.
For now, we show for every deadline its title, its date, the time remaining or if it is done or already due.
If the time remaining is small, we change the color of the text and its style.

I have also removed the greeting activity from the booktamp as it is not useful anymore.


app/src/main/java/com/github/multimatum_team/multimatum/model/DeadlineAdapter.kt
```diff
@@ -0,0 +1,88 @@
...
+import android.graphics.Color
+import android.graphics.Typeface
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import android.widget.BaseAdapter
+import android.widget.TextView
+import com.github.multimatum_team.multimatum.R
+/**
+Class who is used to show the deadline in a clear list.
```
felko, line 14: Nitpick: the docstring should be indented and attached to the class declaration.

app/src/test/java/com/github/multimatum_team/multimatum/DeadlineAdapterTest.kt
```diff
@@ -0,0 +1,124 @@
...
+            Deadline("Number 2",DeadlineState.TODO, LocalDate.now().plusDays(7)),
+            Deadline("Number 3",DeadlineState.DONE, LocalDate.of(2022, 3,30)),
+            Deadline("Number 4",DeadlineState.TODO, LocalDate.of(2022, 3,1)))
+        context = ApplicationProvider.getApplicationContext()
+        adapter = DeadlineAdapter(context!!, list)
+    }
+
+
+    @Test
+    fun testGetCount(){
```
felko, line 40: Nitpick again: you could use Kotlin's ``fun `Explain the test scenario in plain english here`()`` syntax so that all of our tests are consistent.

app/src/main/java/com/github/multimatum_team/multimatum/model/DeadlineAdapter.kt
```diff
@@ -0,0 +1,88 @@
...
+
+        //show the date
+        subtitleTextView.text = "Due the ${deadline.date}"
+        subtitleTextView.setTypeface(null, Typeface.ITALIC)
+
+        //show how much time left or if it is due or done.
+        val detail: String
+        when {
+            deadline.state == DeadlineState.DONE -> {
+                detail = "Done"
```
felko, line 67: Instead of hard coding strings we should use Android resources to enable internationalization eventually.

app/src/main/java/com/github/multimatum_team/multimatum/model/DeadlineAdapter.kt
```diff
@@ -0,0 +1,88 @@
...
+        // Get title element
+        val titleTextView = rowView.findViewById(R.id.deadline_list_title) as TextView
+        // Get subtitle element
+        val subtitleTextView = rowView.findViewById(R.id.deadline_list_subtitle) as TextView
+        // Get detail element
+        val detailTextView = rowView.findViewById(R.id.deadline_list_detail) as TextView
+
+        val deadline = getItem(position) as Deadline
+
+        //show the title
```
felko, line 55: Extreme nitpicking: add a space after the `//` in all your comments.

felko (changes requested):
Everything seems good, the list of deadlines is hardcoded for now but I will take care of fetching the deadline list from Firebase in my PR.

app/src/main/java/com/github/multimatum_team/multimatum/model/DeadlineAdapter.kt
```diff
@@ -0,0 +1,88 @@
...
+        return position.toLong()
+    }
+
+    // Customise the units of the list
+    @SuppressLint( "ViewHolder")
+    override fun getView(position: Int, convertView: View?, parent: ViewGroup): View {
+        // Get view for row item
+        val rowView = inflater.inflate(R.layout.list_item_deadline, parent, false)
+        // Get title element
+        val titleTextView = rowView.findViewById(R.id.deadline_list_title) as TextView
```
ValentinAebi, line 47: You shouldn't have to do such casts: in Kotlin `findViewById` can take a type parameter (at least it worked in the code I wrote):
``` Kotlin
val titleTextView = rowView.findViewById<TextView>(R.id.deadline_list_title)
```

app/src/main/java/com/github/multimatum_team/multimatum/model/DeadlineAdapter.kt
```diff
@@ -0,0 +1,88 @@
...
+        titleTextView.setTypeface(null, Typeface.BOLD)
+
+        // Show the date
+        subtitleTextView.text = context.getString(R.string.DueTheX,deadline.date)
+        subtitleTextView.setTypeface(null, Typeface.ITALIC)
+
+        // Show how much time left or if it is due or done.
+        val detail: String
+        when {
+            deadline.state == DeadlineState.DONE -> {
```
ValentinAebi, line 66: Nitpicking again, but it would clearer (and safer?) to write this in a functional style:
``` Kotlin
val detail = when {
```
and then remove all the `detail =` in the `when` block, e.g. line 69:
``` Kotlin
deadline.isDue -> context.getString(R.string.isAlreadyDue)
```

app/src/main/java/com/github/multimatum_team/multimatum/model/DeadlineAdapter.kt
```diff
@@ -0,0 +1,88 @@
...
+                detail = context.getString(R.string.done)
+            }
+            deadline.isDue -> {
+                detail = context.getString(R.string.isAlreadyDue)
+            }
+            else -> {
+                detail = context.getString(R.string.DueInXDays,deadline.timeRemaining?.days.toString())
```
ValentinAebi, line 76: Is there a particular reason why you are using `?` on `timeRemaining` here and `!!` in the rest of the code block? I would say that if `timeRemaining` is null an exception should be thrown instead of displaying "Due in null days".

app/src/test/java/com/github/multimatum_team/multimatum/DeadlineAdapterTest.kt
```diff
@@ -0,0 +1,124 @@
...
+        Assert.assertEquals(adapter.getItemId(2), 2.toLong())
+        Assert.assertEquals(adapter.getItemId(3), 3.toLong())
+    }
+
+    @Test
+    fun `GetView should show properly the first deadline who is due in 1 day in red`(){
+        val parent = ListView(context)
+        val listItemView: View = adapter.getView(0,null,parent)
+        //check title
+        Assert.assertEquals("Number 1",(listItemView.findViewById(R.id.deadline_list_title) as TextView).text)
```
ValentinAebi, line 65: Same as before, you can pass a type parameter to `findViewById` instead of casting its return value

app/src/main/java/com/github/multimatum_team/multimatum/model/DeadlineAdapter.kt
Matathiasq13824: It could be a possibility but we don't only define detail, we define also it's color and it's type sometimes so I think It will be less clear if I do 2 when

LennyDZ:
I'ld say it's a bad practice to remove things in a PR that is not mean for it. Maybe we should've done a separate PR to remove the greeting Activity.

app/src/main/java/com/github/multimatum_team/multimatum/model/DeadlineAdapter.kt
Matathiasq13824: It is an error, I should have put `!!` everywhere. `timeRemaining` will give null only if the deadline is due. As we check before if it is due, we are sure to not have a null

ValentinAebi (approved):

PR #45, opened 2022-03-17
Testing/dep inject

### LINKED ISSUE #31: Setup dependency injection

ValentinAebi:
I made the setup to use dependency injection with Hilt and Mockito in the project, and I used this to improve the tests of MainSettingsActivity. Despite many attempts, I did not manage to make it work with Mockk, so I used Mockito. I am not sure of whether I am doing something wrong, or Mockk can simply not be used in instrumented tests due to some incompatibilities with configuration of the project. I am going to write a small guide for the use of these tests frameworks in the project.

Closes #31


app/src/main/java/com/github/multimatum_team/multimatum/MultimatumApp.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import android.app.Application
+import dagger.hilt.android.HiltAndroidApp
+
+@HiltAndroidApp
+class MultimatumApp: Application() {
+
+
+
+}
```
LennyDZ, line 11: What is the purpose of this file?

ValentinAebi: Quoting the documentation:
> All apps that use Hilt must contain an [Application](https://developer.android.com/reference/android/app/Application) class that is annotated with @HiltAndroidApp.

I am not exactly sure why this is required, but it is probably because it is the only place where we can put the `@HiltAndroidApp` annotation that triggers code generation.
But you are right, it is not self-explanatory. I am going to add a comment in the file.

app/src/androidTest/java/com/github/multimatum_team/multimatum/MainSettingsActivityTest.kt
```diff
@@ -74,36 +87,51 @@ class MainSettingsActivityTest {
...
+        val mockEditor: SharedPreferences.Editor = mock()
+        `when`(mockEditor.putBoolean(eq(MainSettingsActivity.NOTIF_ENABLED_PREF_KEY), any())).then {
+            assertEquals(expectedFinalNotifEnabled, it.getArgument(1))
+            mockEditor
+        }
+        `when`(mockEditor.putBoolean(eq(MainSettingsActivity.DARK_MODE_PREF_KEY), any())).then {
+            assertEquals(expectedFinalDarkModeEnabled, it.getArgument(1))
+            mockEditor
+        }
+        `when`(mockEditor.apply()).then { /* do nothing */ }
```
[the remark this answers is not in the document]
ValentinAebi: `.then` is used to define the behavior that the mocked object should have when a call as specified in `when` occurs. Here it might work if I remove the `then` (not sure), but I think it is clearer as it is now anyway, because the expected behavior on a call is indeed to "do nothing".

LennyDZ (approved):

felko (approved):
Looks great! Couldn't find anything to add other than what Lenny has already mentionned.

PR #67, opened 2022-03-22
Adding calendar view

macintosh1398:
This new feature (calendar) gives the user the opportunity to add a deadline to a specific date using the new calendar view. It also allows to see the deadlines on a given date using this new interface. The top "Action Bar" was also removed to get more space and to look more like a recent app.

### DIFF
--- a/app/src/main/res/values-night/themes.xml
+++ b/app/src/main/res/values-night/themes.xml
@@ -1,6 +1,6 @@
 <resources xmlns:tools="http://schemas.android.com/tools">
     <!-- Base application theme. -->
-    <style name="Theme.Multimatum" parent="Theme.MaterialComponents.DayNight.DarkActionBar">
+    <style name="Theme.Multimatum" parent="Theme.MaterialComponents.DayNight.NoActionBar">
         <!-- Primary brand color. -->
         <item name="colorPrimary">@color/purple_200</item>
         <item name="colorPrimaryVariant">@color/purple_700</item>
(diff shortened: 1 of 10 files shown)


Voxanimus (commented):
Not sure to understand. The view is working or not ?  cBecaus when i see the dunciton addNewDeadlineCalendar() it seems to do nothing for the moment.
Is the the view display some hardcoded stuff like the list view or not ?

app/src/main/res/values-night/themes.xml
```diff
@@ -1,6 +1,6 @@
 <resources xmlns:tools="http://schemas.android.com/tools">
     <!-- Base application theme. -->
-    <style name="Theme.Multimatum" parent="Theme.MaterialComponents.DayNight.DarkActionBar">
+    <style name="Theme.Multimatum" parent="Theme.MaterialComponents.DayNight.NoActionBar">
```
LennyDZ, line 3: What is this about? Maybe notify this change in the PR.

macintosh1398:
The database doesn't exist yet. So for now, it is not possible to store a deadline that is entered by the user for a specific date, even just a deadline. So now, the deadline can be entered by the user, but it is only temporary displayed (as made for the main screen). The function is not hardcoded (the user can enter a deadline), but it will require some changes when the database will be set up.

Voxanimus (approved):
Sounds fair

LennyDZ (approved):
This seems fine to me.

app/src/main/res/values-night/themes.xml
macintosh1398: Good point! As discussed with members of the team last week, I removed the "Action Bar" (top purple bar) to get more space to look more like a recent app.

PR #69, opened 2022-03-22
refactor notification as static function in a class

LennyDZ:
The function to create notification and notification are now part of DeadlineNotification class as static function.

Also add DeadlineNotificationTest, which is empty yet


app/src/main/java/com/github/multimatum_team/multimatum/DeadlineNotification.kt
```diff
@@ -36,16 +33,18 @@ class DeadlineNotification {
...
-            intent.putExtra("id", id)
+            intent.putExtra("title", deadline.title)
+            intent.putExtra("description", deadline.description)
+            intent.putExtra("id", deadline.id)
 
             //set the receiver as pending intent
-            pendingIntent = PendingIntent.getBroadcast(context, id, intent, PendingIntent.FLAG_IMMUTABLE)
+            pendingIntent = PendingIntent.getBroadcast(context, deadline.id, intent, PendingIntent.FLAG_IMMUTABLE)
+
+            val timeMS: Long = 0 //get milis from deadline.date (LocalDate)
```
ValentinAebi, line 47: What is this line doing? I don't understand the comment.

LennyDZ: This is creates a pending intent of broadcastreceiver that’ll be trigger by the alarm.
But that was already part of the code of last week.

ValentinAebi: I am asking now because I didn't review this last week :) . But I was asking about `val timeMS: Long = 0`. I don't understand what the comment means, and what timeMs is.

LennyDZ: Yeah, it's because timeMs is the time when the notification will be set. So I have to get it from notification date, however we haven't decide yet how it'll be manage (when exactly the notif have to be launch). So I just set it to 0 for now (I think I'll change to currentTime.getMilis in order to make it testable)

ValentinAebi:
I accidentally hit the "Approve" button but I wanted to hit "Comment", so I requested another review from my me. Sorry for the confusion!

app/src/main/java/com/github/multimatum_team/multimatum/DeadlineNotification.kt
ValentinAebi: OK, thanks for the explanation. Maybe you can put a comment like "Temporary"?

PR #71, opened 2022-03-23
Feature/procrastination detector

### LINKED ISSUE #60: Implement a basic system to detect whether the user is sitting at his/her desk working or doing something else

ValentinAebi:
I implemented a very basic version of the procrastination/distraction detection feature. The idea is to detect when the phone is moving and, if the user has a deadline soon, to notify him/her because this means that s-he is probably not working at the moment. Currently, detection works only if the app is open and does not take into account the imminence of a deadline, this should be extended later.
I tried to do tests for this feature, but I did not manage to make tests work (at least for now), I will probably try again next week. So unfortunately this PR will decrease coverage, I am sorry about that.

Closes #60


app/src/main/java/com/github/multimatum_team/multimatum/procrastination_detector/ProcrastinationDetectorService.kt
```diff
@@ -0,0 +1,82 @@
...
+import android.hardware.SensorManager
+import android.os.IBinder
+import android.widget.Toast
+import com.github.multimatum_team.multimatum.R
+import dagger.hilt.android.AndroidEntryPoint
+import javax.inject.Inject
+import kotlin.math.abs
+
+@AndroidEntryPoint
+class ProcrastinationDetectorService : Service(), SensorEventListener {
```
Matathiasq13824, line 17: It should be nice to add a description of the class and multiples comments before the function to explain what they do

app/src/main/java/com/github/multimatum_team/multimatum/MainSettingsActivity.kt
```diff
@@ -48,6 +96,15 @@ class MainSettingsActivity : AppCompatActivity() {
...
+         * Key for the storage of whether procrastination fighter is enabled
+         */
+        const val PROCRASTINATION_FIGHTER_ENABLED_PREF_KEY =
+            "com.github.multimatum_team.multimatum.MainSettingsActivity.ProcrastinationFighterEnabled"
+
+        private const val BUTTON_ALPHA_NORMAL = 1F
+        private const val BUTTON_ALPHA_DISABLED = 0.5F
     }
 
 }
```
macintosh1398, line 110: Looks good, very modular

app/src/main/java/com/github/multimatum_team/multimatum/procrastination_detector/ProcrastinationDetectorService.kt
```diff
@@ -0,0 +1,82 @@
...
+        toast("Procrastination fighter is disabled")
+    }
+
+    override fun onAccuracyChanged(sensor: Sensor?, accuracy: Int) { /* Nothing to do */ }
+
+    override fun onSensorChanged(event: SensorEvent?) {
+        val currentTime = System.currentTimeMillis()
+        if (currentTime >= lastDetection + MIN_PERIOD_BETWEEN_NOTIF){
+            val currentPosition = event!!.values.toTypedArray()
+            if (lastPosition != null && l1Distance(currentPosition, lastPosition!!) > MOVE_DETECTION_THRESHOLD) {
```
Matathiasq13824, line 51: As I see, a Toast appear when you move from where you where before. However, a distance of 0.1 is really short. You could do a movment like this when you add a deadline in Mulmatum, which will render the feature annoying even in the app.

macintosh1398 (approved):
Good work, I approve the PR. But it is absolutely necessary to add tests in the coming weeks for ProcrastinationDetectorService.kt

app/src/main/java/com/github/multimatum_team/multimatum/MainSettingsActivity.kt
```diff
@@ -1,40 +1,88 @@
...
+            if (sensorPresent) BUTTON_ALPHA_NORMAL else BUTTON_ALPHA_DISABLED
+        if (!sensorPresent) {
+            procrastinationDetectEnabledButton.text = applicationContext.getString(
+                R.string.missing_sensor_msg, procrastinationDetectEnabledButton.text
+            )
+        }
+        procrastinationDetectEnabledButton.isClickable = sensorPresent
+    }
+
+    private fun setWidgetListeners() {
```
Matathiasq13824, line 46: Add comments to make your code more understandable

Matathiasq13824 (changes requested):
It's a pity to not have the tests for it. We have to plan them for the next sprint.
You should also add comments to make easier the reading of your code.
And, last nitpick, you should correct the warning in the string.xml

ValentinAebi:
@Matathiasq13824 I know for the tests and as I said I am sorry for that. They are already planned for next week. About the warning, are you talking about the ellipsis? If yes I won't correct it because I am not the author of this line, and I don't know how to make it better.

app/src/main/java/com/github/multimatum_team/multimatum/procrastination_detector/ProcrastinationDetectorService.kt
ValentinAebi: 0.1 is a threshold that I determined experimentally. It has no unit because it refers to the values given by the sensor, which don't have units. And yes, currently it displays the toast even if the movement comes from setting a deadline in the app. Eventually the goal will probably be to avoid this behavior, e.g. by stopping the service when the app is used. But the current implementation is incomplete and a bit experimental, and does not support this complex behavior. But you are right about the fact that this is confusing, I'll add a comment.

Matathiasq13824 (approved):

app/src/main/java/com/github/multimatum_team/multimatum/procrastination_detector/ProcrastinationDetectorService.kt
Matathiasq13824: Okay, thanks for the precisions :)

PR #80, opened 2022-03-24
Refactor+test/notifications

LennyDZ:
Refactor notification as a class that provide static functions.

Add test for ReminderBroadcastReceiver, which handle alarm to launch notifications


app/src/androidTest/java/com/github/multimatum_team/multimatum/DeadlineNotificationTest.kt
```diff
@@ -0,0 +1,21 @@
+package com.github.multimatum_team.multimatum
```
felko, line 1: What's the reason behind this test? I assume it's for testing the CI.
If this is true, and if you think you're not going to be able to make tests work tonight, please delete this file.

app/src/main/java/com/github/multimatum_team/multimatum/DeadlineNotification.kt
```diff
@@ -0,0 +1,55 @@
...
+import android.app.AlarmManager
+import android.app.NotificationChannel
+import android.app.NotificationManager
+import android.app.PendingIntent
+import android.content.Context
+import android.content.Intent
+import androidx.appcompat.app.AppCompatActivity
+import com.github.multimatum_team.multimatum.model.Deadline
+
+class DeadlineNotification {
```
felko, line 12: Please add documentation for the whole class

app/src/main/java/com/github/multimatum_team/multimatum/DeadlineNotification.kt
```diff
@@ -0,0 +1,55 @@
...
+    companion object {
+        private lateinit var alarmManager: AlarmManager
+        private lateinit var pendingIntent: PendingIntent
+        /*
+        Create a notification channel for reminder notifications
+        Creating an existing notification channel with its original values performs no operation,
+        so it's safe to call this code when starting an app.
+        */
```
felko, line 20: Documentation is written between `/**` and `*/` (notice the double `*`), please use this convention as well (the doc should appear in green in android studio)

app/src/main/java/com/github/multimatum_team/multimatum/DeadlineNotification.kt
```diff
@@ -0,0 +1,55 @@
...
+
+            channel.description=description
+
+            notificationManager.createNotificationChannel(channel)
+        }
+
+        /*
+        Set a notification that will be triggered in a given time in ms.
+        you can pass a title/description and Id in parameter
+        */
```
felko, line 35: Same here

app/src/main/java/com/github/multimatum_team/multimatum/model/Deadline.kt
```diff
@@ -36,7 +36,9 @@ enum class DeadlineState {
 data class Deadline(
     val title: String,
     val state: DeadlineState,
-    val date: LocalDate
+    val date: LocalDate,
+    val description: String = "",
+    val id: Int = 0
```
felko, line 41: Firebase generates random IDs automatically, but they're represented as strings.
Please make it so that `id` is of type `String`.
Defining default values for the new fields is okay for now if you want to get things working quickly, but is ultimately going to cause us some problems, we'll have to remove the default values and fix code breakage in the next sprint.

felko (changes requested):

felko (approved):

Voxanimus (approved):
There is a lot of test that remain in comments, it will be good to make them work.
But the remaining part seems good.

PR #93, opened 2022-03-28
Add clock service to enable reproducible testing

### LINKED ISSUE #92: Create clock service to allow reproducible testing
Define a `ClockService` interface providing a way to get the current date/time. A singleton will be injected into all classes that need access to time (mainly `Deadline`). Currently, there was an attempt of using `mockk` to mock `Clock.getSystemUTC()` but this turns out to be unreliable. With the clock service, we can inject a mock clock in the unit tests to make them reproducible.

felko:
This PR implements a `ClockService` interface that enables one to get the clock and (local) date. The real implementation of the service is provided by the `SystemClockService` class, giving the actual date at which the code is executed. The purpose of this service is to be able to use dependency injection to inject a mock clock in time-sensitive code. This lets us write reproducible tests that won't fail when executed at certain times.

Closes #92

### DIFF
--- a/app/src/main/java/com/github/multimatum_team/multimatum/DependenciesProvider.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/DependenciesProvider.kt
@@ -9,6 +9,9 @@ import com.github.multimatum_team.multimatum.model.DeadlineState
 import com.github.multimatum_team.multimatum.repository.DeadlineRepository
 import com.github.multimatum_team.multimatum.repository.FirebaseDeadlineRepository
 import android.hardware.SensorManager
+import android.os.SystemClock
+import com.github.multimatum_team.multimatum.service.ClockService
+import com.github.multimatum_team.multimatum.service.SystemClockService
 import dagger.Module
 import dagger.Provides
 import dagger.hilt.InstallIn
@@ -30,6 +33,14 @@ object DependenciesProvider {
     @Provides
     fun provideSensorManager(@ApplicationContext applicationContext: Context): SensorManager =
         applicationContext.getSystemService(SENSOR_SERVICE) as SensorManager
+}
+
+@Module
+@InstallIn(SingletonComponent::class)
+object ClockModule {
+    @Provides
+    fun provideClockService(): ClockService =
+        SystemClockService()
 }
 
 @Module
--- a/app/src/main/java/com/github/multimatum_team/multimatum/model/DeadlineAdapter.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/model/DeadlineAdapter.kt
@@ -16,7 +16,6 @@ import dagger.hilt.InstallIn
 import dagger.hilt.android.EntryPointAccessors
 import dagger.hilt.components.SingletonComponent
 import java.time.Period
-import javax.inject.Inject
 
 @EntryPoint
 @InstallIn(SingletonComponent::class)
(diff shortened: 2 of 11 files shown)


app/src/main/java/com/github/multimatum_team/multimatum/DependenciesProvider.kt
```diff
@@ -32,6 +35,14 @@ object DependenciesProvider {
         applicationContext.getSystemService(SENSOR_SERVICE) as SensorManager
 }
 
+@Module
+@InstallIn(SingletonComponent::class)
+object ClockModule {
```
Matathiasq13824, line 40: Just a question: why you separate it in another module ? It is to have a better code ?

app/src/main/java/com/github/multimatum_team/multimatum/model/DeadlineAdapter.kt
```diff
@@ -10,19 +10,37 @@ import android.view.ViewGroup
...
+    fun provideClockService(): ClockService
+}
 
 /**
 Class which is used to show the deadline in a clear list.
 Based on the tutorial:
 https://www.raywenderlich.com/155-android-listview-tutorial-with-kotlin
  */
-class DeadlineAdapter(private val context: Context) : BaseAdapter() {
+class DeadlineAdapter @Inject constructor(private val context: Context) : BaseAdapter() {
```
Matathiasq13824, line 32: I didn't see in the code somewhere where this injection is used. Did I just miss it ?

Matathiasq13824 (commented):
Great code that will be very useful during the next weeks. I just need some explanations of some part of the code before approving

app/src/main/java/com/github/multimatum_team/multimatum/DependenciesProvider.kt
felko: If I put the clock service in our DependenciesProvider module, I'd have to mock the sensor and shared preferences as well in the tests, even if the tested code doesn't use any of those. The way I understand it, the provided dependencies in a module should be coupled together (e.g. if I have another dependency related to repositories, I'd put it in the repository module).

app/src/main/java/com/github/multimatum_team/multimatum/model/DeadlineAdapter.kt
felko: Good catch, that was from an earlier attempt to inject the clock service. It seems to work without it.
Though dependency injection _is_ used in this class (the clockService is injected). Since Hilt doesn't support injecting dependencies into adapters (it only supports a few android classes out of the box), I have to define an entry point and maually inject the clock service.

app/src/main/java/com/github/multimatum_team/multimatum/model/DeadlineAdapter.kt
```diff
@@ -75,17 +92,18 @@ class DeadlineAdapter(private val context: Context) : BaseAdapter() {
...
+                    context.getString(R.string.DueInXDays, timeRemaining.days.toString())
                 // If the remaining days is too small, put them in red or orange
-                if ((deadline.timeRemaining!!.days) < URGENT) {
+                if (timeRemaining.days < URGENT) {
                     detailTextView.setTextColor(Color.RED)
                     detailTextView.setTypeface(null, Typeface.BOLD)
-                } else if ((deadline.timeRemaining!!.days) < PRESSING) {
+                } else if (timeRemaining.days < PRESSING) {
                     detailTextView.setTextColor(Color.rgb(255, 165, 0))
                     detailTextView.setTypeface(null, Typeface.BOLD)
```
LennyDZ, line 108: Maybe we could add PRESSING and URGENT, as deadline state in the deadline model. Then we could set a function getState to deadline class and use this instead of constant defined in the adapter (it seems bad to me to define constant in an adapter)

felko: The issue is that you'd have to update all deadlines from the database every day to change the states to pressing or urgent. I think pressing and urgent are not really deadline states, it just controls how they are displayed on the view.

LennyDZ (approved):

PR #94, opened 2022-03-29
Feature/add deadline button

Matathiasq13824:
Added an activity to add deadline on the list using a DatePicker.

### DIFF
--- a/app/src/androidTest/java/com/github/multimatum_team/multimatum/MainActivityTest.kt
+++ b/app/src/androidTest/java/com/github/multimatum_team/multimatum/MainActivityTest.kt
@@ -117,6 +117,17 @@ class MainActivityTest {
     }
 
     @Test
+    fun goToAddDeadlineActivity() {
+        onView(withId(R.id.main_go_to_add_deadline)).perform(ViewActions.click())
+        Intents.intended(
+            allOf(
+                hasComponent(AddDeadlineActivity::class.java.name),
+                toPackage("com.github.multimatum_team.multimatum")
+            )
+        )
+    }
+
+    @Test
     fun buttonOpensQrCodeReader() {
         onView(withId(R.id.goToQrCodeReader)).perform(ViewActions.click())
         Intents.intended(
--- a/app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
@@ -15,7 +15,7 @@ import java.time.LocalDate
 import javax.inject.Inject
 
 /**
- *  Activity who create a deadline using a DatePick
+ *  Activity who create a deadline using a DatePicker
  */
 @AndroidEntryPoint
 class AddDeadlineActivity : AppCompatActivity() {
--- a/app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
@@ -25,9 +25,6 @@ class AddDeadlineActivity : AppCompatActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_add_deadline)
-
-        val datePicker = findViewById<DatePicker>(R.id.add_deadline_date_picker)
-        datePicker.init(2022, 1, 1, null)
     }
 
     fun addDeadline(view: View) {
@@ -41,13 +38,13 @@ class AddDeadlineActivity : AppCompatActivity() {
 
         // Check if the title is not empty
         if (titleDeadline == "") {
-            Toast.makeText(this, "Enter a title", Toast.LENGTH_SHORT).show()
+            Toast.makeText(this, getString(R.string.enter_a_title), Toast.LENGTH_SHORT).show()
         } else {
             // Add the deadline
             val deadline = Deadline(titleDeadline, DeadlineState.TODO, dateDeadline)
             runBlocking { deadlineRepository.put(deadline) }
 
-            Toast.makeText(this, "Deadline created.", Toast.LENGTH_SHORT).show()
+            Toast.makeText(this, getString(R.string.deadline_created), Toast.LENGTH_SHORT).show()
 
             // Reset the text input for future use
             editText.text = ""
--- a/app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
@@ -19,12 +19,20 @@ import javax.inject.Inject
  */
 @AndroidEntryPoint
 class AddDeadlineActivity : AppCompatActivity() {
+
+    companion object {
+        // Temporary value that will be link later to the actual day
+        private val initDate : LocalDate = LocalDate.of(2022,1,1)
+    }
+
     @Inject
     lateinit var deadlineRepository: DeadlineRepository
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_add_deadline)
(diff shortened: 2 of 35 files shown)


app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
```diff
@@ -0,0 +1,60 @@
...
+class AddDeadlineActivity : AppCompatActivity() {
+    @Inject
+    lateinit var deadlineRepository: DeadlineRepository
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_add_deadline)
+
+        val datePicker = findViewById<DatePicker>(R.id.add_deadline_date_picker)
+        datePicker.init(2022, 1, 1, null)
```
ValentinAebi, line 30: This is just a question, but why do you initialize the date picker to 1.01.2022? And I think that this "initialization date" should be a constant.

app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
```diff
@@ -0,0 +1,60 @@
...
+        val datePicker = findViewById<DatePicker>(R.id.add_deadline_date_picker)
+        val editText = findViewById<TextView>(R.id.add_deadline_select_title)
+
+        // Getting the entered text and the selected date
+        val titleDeadline = editText.text.toString()
+        val dateDeadline = LocalDate.of(datePicker.year, datePicker.month, datePicker.dayOfMonth)
+
+        // Check if the title is not empty
+        if (titleDeadline == "") {
+            Toast.makeText(this, "Enter a title", Toast.LENGTH_SHORT).show()
```
ValentinAebi, line 44: "Enter a title" should be a resource

app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
```diff
@@ -0,0 +1,60 @@
...
+        } else {
+            // Add the deadline
+            val deadline = Deadline(titleDeadline, DeadlineState.TODO, dateDeadline)
+            runBlocking { deadlineRepository.put(deadline) }
+
+            Toast.makeText(this, "Deadline created.", Toast.LENGTH_SHORT).show()
```
ValentinAebi, line 50: "Deadline created" should be a resource too

app/src/test/java/com/github/multimatum_team/multimatum/AddDeadlineTest.kt
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Class test for AddDeadlineActivity
+ */
+@UninstallModules(RepositoryModule::class)
+@HiltAndroidTest
+@RunWith(AndroidJUnit4::class)
+class AddDeadlineTest {
+
+    @get:Rule
+    var hiltRule = HiltAndroidRule(this)
```
ValentinAebi, line 37: The Hilt Rule should always be executed first. From what I understood, the rules (i.e. the things annotated with @Rule) execute in the order they are declared, so this should not be an issue for now. However, it would be safer to enforce the ordering:
```Kotlin
@get:Rule(order = 0)
    var hiltRule = HiltAndroidRule(this)
```
And then orders 1, 2, etc. for the following rules.

app/src/test/java/com/github/multimatum_team/multimatum/AddDeadlineTest.kt
```diff
@@ -0,0 +1,83 @@
...
+    fun `The button should add a deadline given it's title and its date`() {
+        Espresso.onView(ViewMatchers.withId(R.id.add_deadline_select_title))
+            .perform(ViewActions.replaceText("Test 1"))
+        Espresso.closeSoftKeyboard()
+        Espresso.onView(ViewMatchers.withId(R.id.add_deadline_date_picker))
+            .perform(PickerActions.setDate(2022, 5, 8))
+        Espresso.onView(ViewMatchers.withId(R.id.add_deadline_button)).perform(ViewActions.click())
+        MatcherAssert.assertThat(
+            ShadowToast.getTextOfLatestToast(),
+            CoreMatchers.equalTo("Deadline created.")
```
ValentinAebi, line 69: If you turn "Deadline created" into a resource as I suggested, you can replace it here too so that it is easier to change the message. The same holds for "Enter a title" in the previous test.

app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
Matathiasq13824: This could perfectly be changed to initiate it on the actual date however I was waiting the implementation of the ClockService interface before doing so

Matathiasq13824: My bad, it is not necessary. Like CalendarView, we don't need to initiate it.

app/src/test/java/com/github/multimatum_team/multimatum/AddDeadlineTest.kt
Matathiasq13824: Unfortunatly, you can't use the resource of string.xml in a test. It is not possible to use `getString()` unless you use a `@Mock`. I think however it is a good idea to keep it like that to check that the code correctly recuperate the resource.

ValentinAebi: Does `R.string.name_of_the_string_resource` not work in tests?

ValentinAebi: Oh OK I think I see the problem. You can use `applicationContext.getString(...)`

ValentinAebi (approved):

app/src/test/java/com/github/multimatum_team/multimatum/AddDeadlineTest.kt
Matathiasq13824: Unfortunatly, `applicationContext.getString(...)` doesn't work either

ValentinAebi: You should be able to use `RuntimeEnvironment.getApplication().applicationContext.getString(R.string.<your_resource>)` (hopefully it works this time...)

Matathiasq13824: This works ! Thanks !

app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
Matathiasq13824: Edit after some modification: 
It seems to be necessary to initiate it if we don't want to have some errors during the tests.
As I have not enough time, I will link this date next sprint to init the datePicker at the actual date

app/src/androidTest/java/com/github/multimatum_team/multimatum/MainActivityTest.kt
```diff
@@ -116,6 +116,17 @@ class MainActivityTest {
...
     }
 
+    @Test
+    fun goToAddDeadlineActivity() {
+        onView(withId(R.id.main_go_to_add_deadline)).perform(ViewActions.click())
+        Intents.intended(
+            allOf(
+                hasComponent(AddDeadlineActivity::class.java.name),
+                toPackage("com.github.multimatum_team.multimatum")
+            )
```
Voxanimus, line 126: the allOf() function is not necessary you can just do Intents.intended(hasComponent(AddDedlineActivity::class.java.name)
The best will be clearer and won't have useless line

Voxanimus (approved):

app/src/androidTest/java/com/github/multimatum_team/multimatum/MainActivityTest.kt
Matathiasq13824: But this will remove the check of the correct package by doing so I think

PR #95, opened 2022-03-29
Identify deadlines by ID

### LINKED ISSUE #90: Create function to get deadline by Id in the adapter
This function will be use to retrieve deadline corresponding to notification

felko:
This PR changes the model from a list of deadlines to a map associating IDs to deadlines. This lets us augment the repository interface with new methods, namely `fetch` (get a single deadline from the repository) `delete` and `modify`, which will be useful to implement core features later.
The viewmodel and the adapter were updated accordingly.
Other minor refactors were done, for example the viewmodel is now injected in relevant activities instead of being manually instantiated.

Closes #90

### DIFF
--- a/app/src/main/java/com/github/multimatum_team/multimatum/MainActivity.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/MainActivity.kt
@@ -42,7 +42,7 @@ class MainActivity : AppCompatActivity() {
         val adapter = DeadlineAdapter(this)
 
         listView.adapter = adapter
-        viewModel.deadlines.observe(this) { deadlines ->
+        viewModel.getDeadlines().observe(this) { deadlines ->
             Log.d("deadlines", deadlines.toString())
             adapter.setDeadlines(deadlines)
         }
--- a/app/src/main/java/com/github/multimatum_team/multimatum/MainActivity.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/MainActivity.kt
@@ -49,8 +49,8 @@ class MainActivity : AppCompatActivity() {
 
         // Set when you maintain your finger on an item of the list, launch the detail activity
         listView.setOnItemLongClickListener { _, _, position, _ ->
-            val selectedDeadline = adapter.getItem(position)
-            val detailIntent = DeadlineDetailsActivity.newIntent(this, selectedDeadline)
+            val (id, selectedDeadline) = adapter.getItem(position)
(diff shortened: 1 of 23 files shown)


app/src/main/java/com/github/multimatum_team/multimatum/MainActivity.kt
```diff
@@ -26,7 +27,8 @@ class MainActivity : AppCompatActivity() {
     @Inject
     lateinit var deadlineRepository: DeadlineRepository
 
-    private lateinit var viewModel: DeadlineListViewModel
+    private val viewModel: DeadlineListViewModel by viewModels()
```
Matathiasq13824, line 30: Why is the `by viewModel()` necessary ?

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseDeadlineRepository.kt
```diff
@@ -54,41 +52,66 @@ class FirebaseDeadlineRepository @Inject constructor() : DeadlineRepository {
...
 
     /**
      * Insert new deadline in the database.
      */
-    override suspend fun put(deadline: Deadline) {
+    override suspend fun put(deadline: Deadline): DeadlineID =
         deadlinesRef
             .add(serializeDeadline(deadline))
             .await()
+            .id
```
Matathiasq13824, line 78: How Firebase create the id ? Can it create twice the same id if there is enough deadline ?

Matathiasq13824 (commented):
Clean and comprehensible modification. I just need some explaination and it should be good.
However, I see that the coverage got worse because of the modifications. It is possible to make some test to improve it ?
Also, I know that the warnings don't come from your part but it could be nice to, at least, remove the ones who are in the gradle.

LennyDZ (approved):

app/src/main/java/com/github/multimatum_team/multimatum/MainActivity.kt
felko: It is not strictly necessary for this PR but since I was working on the viewmodel I refactored the activities to inject the viewmodel.
This will allow us to test the activities by injecting a mock viewmodel.

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseDeadlineRepository.kt
felko: No the generated ID is guaranteed to be unique.

Matathiasq13824 (approved):

PR #98, opened 2022-03-30
refactoring DeadLine

LennyDZ:
Make DeadLine attribute date become datetime.

### DIFF
--- a/app/src/main/java/com/github/multimatum_team/multimatum/model/DeadlineAdapter.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/model/DeadlineAdapter.kt
@@ -8,6 +8,7 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.widget.BaseAdapter
+import android.widget.Chronometer
 import android.widget.TextView
 import com.github.multimatum_team.multimatum.R
 import com.github.multimatum_team.multimatum.repository.DeadlineID
@@ -17,6 +18,7 @@ import dagger.hilt.InstallIn
 import dagger.hilt.android.EntryPointAccessors
 import dagger.hilt.components.SingletonComponent
 import java.time.Period
+import java.time.temporal.ChronoUnit
 import java.util.*
 
 @EntryPoint
@@ -49,7 +51,7 @@ class DeadlineAdapter(private val context: Context) : BaseAdapter() {
 
     fun setDeadlines(deadlines: Map<DeadlineID, Deadline>) {
         dataSource = deadlines.toSortedMap { id1, id2 ->
-            compareValues(deadlines[id1]!!.date, deadlines[id2]!!.date)
+            compareValues(deadlines[id1]!!.dateTime, deadlines[id2]!!.dateTime)
         }
         notifyDataSetChanged()
     }
@@ -85,7 +87,7 @@ class DeadlineAdapter(private val context: Context) : BaseAdapter() {
         titleTextView.setTypeface(null, Typeface.BOLD)
 
         // Show the date
-        subtitleTextView.text = context.getString(R.string.DueTheX, deadline.date)
+        subtitleTextView.text = context.getString(R.string.DueTheX, deadline.dateTime)
         subtitleTextView.setTypeface(null, Typeface.ITALIC)
 
         // Show how much time left or if it is due or done.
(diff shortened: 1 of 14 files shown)


app/src/test/java/com/github/multimatum_team/multimatum/DeadlineDetailsTest.kt
```diff
@@ -9,28 +9,54 @@ import androidx.test.espresso.matcher.ViewMatchers.withText
...
+        hiltRule.inject()
+    }
 
     @Test
     fun `Given a deadline not yet due or done, the activity should display it`() {
         val intent = DeadlineDetailsActivity.newIntent(
             ApplicationProvider.getApplicationContext(),
             "1",
-            Deadline("Test 1", DeadlineState.TODO, LocalDate.now().plusDays(7))
+            Deadline("Test 1", DeadlineState.TODO, clockService.now().plusDays(7))
```
ValentinAebi, line 54: I am a bit confused about how this test (and more generally all the tests using `clockService.now()`) work. Do they really use the time at which the test is run, or is there some mocking to avoid that? Because in the former case the tests may pass or fail depending on when they are run, which is not acceptable.

app/src/test/java/com/github/multimatum_team/multimatum/util/MockClockService.kt
```diff
@@ -1,19 +1,16 @@
...
-import java.time.LocalDate
-import java.time.ZoneId
+import java.time.*
 
 /**
  * A mock class to simulate a fixed clock.
  * This lets us write more reproducible tests.
  */
-class MockClockService(private var date: LocalDate) : ClockService {
+class MockClockService(private var dateTime: LocalDateTime) : ClockService {
```
ValentinAebi, line 10: I think that `dateTime` should be a val, not a var

app/src/main/java/com/github/multimatum_team/multimatum/model/DeadlineAdapter.kt
```diff
@@ -96,18 +98,18 @@ class DeadlineAdapter(private val context: Context) : BaseAdapter() {
...
             }
             else -> {
-                val timeRemaining = Period.between(clockService.now(), deadline.date)
+                val timeRemaining = clockService.now().until(deadline.dateTime, ChronoUnit.DAYS)
                 detail =
-                    context.getString(R.string.DueInXDays, timeRemaining.days.toString())
+                    context.getString(R.string.DueInXDays, timeRemaining.toString())
                 // If the remaining days is too small, put them in red or orange
-                if (timeRemaining.days < URGENT) {
+                if (timeRemaining < URGENT) {
```
ValentinAebi, line 109: These constants (`URGENT` and `PRESSING`) should be renamed, because comparing a time and "urgent" does not make much sense conceptually. `URGENT` could e.g. become `URGENT_THRESHOLD_DAYS`. `timeRemaining` could also be renamed to `daysRemaining` (or similar) for clarity.

app/src/test/java/com/github/multimatum_team/multimatum/DeadlineDetailsTest.kt
LennyDZ: there's some mocking, it's the purpose of the existence of clockService. It allows us to override now() function to get a special fixed time for tests.

app/src/test/java/com/github/multimatum_team/multimatum/util/MockClockService.kt
LennyDZ: That's none of my business, I just change the type here. But I agree with you, I'll change that.

app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
```diff
@@ -22,7 +23,7 @@ class AddDeadlineActivity : AppCompatActivity() {
 
     companion object {
         // Temporary value that will be link later to the actual day
-        private val initDate : LocalDate = LocalDate.of(2022,1,1)
+        private val initDate : LocalDateTime = LocalDateTime.of(2022,1,1, 0, 0)
```
macintosh1398, line 26: Ctrl+alt+L ;)

macintosh1398 (approved):
You got my approval.

Voxanimus (commented):
what is the concrete difference between localDate and localDateTime ? I'm not sur to understand it

felko:
> what is the concrete difference between localDate and localDateTime ? I'm not sur to understand it

`LocalDate` is YYYY/MM/DD and `LocalDateTime` is YYYY/MM/DD HH:MM:SS.
This will allow us to set the deadline to a specific time of day instead of just the day.

Voxanimus (approved):
With last comment it's more clear to me.
I approve it

ValentinAebi (approved):

PR #104, opened 2022-03-30
Testing/procrastination detector

### LINKED ISSUE #88: Tests for movement detector

ValentinAebi:
Tests on ProcrastinationDetectorService. Most of the tests consist in scenarios where fake events are given to the service and its responses are checked. Tests are built using ServiceController from Robolectric. They also use Hilt and Mockito, as well as some custom reflection in a case where it is needed to bypass immutability.

Closes #88


Matathiasq13824 (changes requested):
Great increase of coverage ! However, it could be nice to add some comments in the functions of `ProcrastinationDetectorService` if someone else have to modify it because some of them are not very inderstandable.

app/src/main/java/com/github/multimatum_team/multimatum/service/ProcrastinationDetectorService.kt
```diff
@@ -6,6 +6,7 @@ import android.hardware.Sensor
 import android.hardware.SensorEvent
```
felko, line 6: Please format this file (ctrl+alt+L)

app/src/test/java/com/github/multimatum_team/multimatum/service/ProcrastinationDetectorServiceTest.kt
```diff
@@ -0,0 +1,249 @@
+package com.github.multimatum_team.multimatum.service
```
felko, line 1: This file could be formatted as well

app/src/test/java/com/github/multimatum_team/multimatum/service/ProcrastinationDetectorServiceTest.kt
```diff
@@ -0,0 +1,249 @@
...
+            0f, 0f, 0f, 2_000_000_000,                  // 1st event
+            TINY_CHANGE, TINY_CHANGE, TINY_CHANGE, 5_000_000_000,  // 2nd event
+            mockSensorEvent, service
+        )
+        assertThatLastToastTextWas(FAKE_TOAST_TEXT)
+        controller.destroy()
+    }
+
+    @Test
+    fun on_bind_should_return_a_binder_that_is_able_to_provide_the_bound_service(){
```
felko, line 124: Nitpick: you can use the ``fun `onBinds should return a binder...`() { ... }`` syntax. Oddly this syntax is not supported in androidTests but it does in regular unit tests.

felko (changes requested):

app/src/test/java/com/github/multimatum_team/multimatum/service/ProcrastinationDetectorServiceTest.kt
ValentinAebi: Actually I already had problems with this syntax where it appeared as `onBind%20should%20return%20a%20binder` (or similar) on the test results, so I prefer to keep it with underscores

felko: I see, then we'll probably have to refactor the other tests to use the normal syntax everywhere

felko (approved):

app/src/test/java/com/github/multimatum_team/multimatum/service/ProcrastinationDetectorServiceTest.kt
ValentinAebi: I don't think that it is worth the effort. I tried running all the tests and did not notice these %20 this time (actually when I noticed them it was in Intellij for another project, so the problem might not even exist on Android Studio, I just stick to underscore notation because I think that it is readable enough and avoid this kind of bugs in case they may arise somewhere). And it is no big deal if we use different conventions for the names of tests I would say.

PR #115, opened 2022-04-01
Fix datetime not being correctly serialized into firebase

felko:
This PR fixes a bug that caused the deadline dates to be inserted into the Firebase database as is, without being properly serialized. As a consequence, we couldn't add deadlines using neither the calendar or the add deadline button.
This wouldn't happen if we had proper tests for the firebase repository, we should find a way to use the firebase emulator suite.


ValentinAebi (approved):
This PR fixes a critical bug, so we must merge it as soon as possible. But we will need to investigate later why coverage is so low...

Matathiasq13824 (approved):
A change of 0,5% in the coverage not enough to not approve this fix

PR #123, opened 2022-04-04
Modified AddDeadlineActivity to add selection of time

Matathiasq13824:
After the refactoring of Deadline modifying LocalDate to LocalDate time, it was necessary to change AddDeadlineActivity to allow the user to select a time for their deadline. As the DatePicker wasn't enough, I have added a TimePicker and put the two of them in a Dialog.

### DIFF
--- a/app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
@@ -11,12 +11,12 @@ import androidx.activity.viewModels
 import androidx.appcompat.app.AppCompatActivity
 import com.github.multimatum_team.multimatum.model.Deadline
 import com.github.multimatum_team.multimatum.model.DeadlineState
+import com.github.multimatum_team.multimatum.service.ClockService
 import com.github.multimatum_team.multimatum.viewmodel.DeadlineListViewModel
 import dagger.hilt.android.AndroidEntryPoint
-import java.time.Instant
 import java.time.LocalDateTime
-import java.time.ZoneId
 import java.time.temporal.ChronoUnit
+import javax.inject.Inject
 
 
 /**
@@ -25,8 +25,11 @@ import java.time.temporal.ChronoUnit
 @AndroidEntryPoint
 class AddDeadlineActivity : AppCompatActivity() {
 
-    private var selectedDate: LocalDateTime =
-        Instant.now().atZone(ZoneId.systemDefault()).toLocalDateTime().truncatedTo(ChronoUnit.HOURS)
+    @Inject
+    lateinit var clockService: ClockService
+
+    private lateinit var selectedDate: LocalDateTime
+
 
     private val viewModel: DeadlineListViewModel by viewModels()
     private lateinit var textDate: TextView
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
```diff
@@ -1,57 +1,103 @@
...
         setContentView(R.layout.activity_add_deadline)
-        val datePicker = findViewById<DatePicker>(R.id.add_deadline_date_picker)
-        datePicker.init(initDate.year,initDate.month.value,initDate.dayOfMonth, null)
+        textDate = findViewById(R.id.add_deadline_text_date)
+        textTime = findViewById(R.id.add_deadline_text_time)
+        textDate.text = selectedDate.toLocalDate().toString()
+        textTime.text = selectedDate.toLocalTime().toString()
+    }
+
+    //Setup a DatePickerDialog that will select a date for the deadline and show it
```
LennyDZ, line 44: Here you should use proper comment format for functions (/** ... */) or make it a private function.

app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
```diff
@@ -1,57 +1,103 @@
...
+
+        // Prepare the datePickerDialog
+        val datePickerDialog = DatePickerDialog(this, dateSetListener,
+            selectedDate.year, selectedDate.month.ordinal, selectedDate.dayOfMonth)
+
+        // Show the Dialog on the screen
+        datePickerDialog.show()
+    }
+
+    //Setup a TimePickerDialog that will select a time for the deadline and show it
```
LennyDZ, line 66: same here.

app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
```diff
@@ -1,57 +1,103 @@
...
+            }
+
+        // Prepare the datePickerDialog
+        val timePickerDialog = TimePickerDialog(this, timeSetListener,
+            selectedDate.hour, selectedDate.minute, true)
+
+        // Show the Dialog on the screen
+        timePickerDialog.show()
+
     }
```
LennyDZ, line 86: Here you could also add some description to the function

app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
```diff
@@ -1,57 +1,103 @@
...
+ *  Activity who create a deadline using a DatePickerDialog and a TimePickerDialog
  */
 @AndroidEntryPoint
 class AddDeadlineActivity : AppCompatActivity() {
 
-    companion object {
-        // Temporary value that will be link later to the actual day
-        private val initDate : LocalDateTime = LocalDateTime.of(2022,1,1, 0, 0)
-    }
+    private var selectedDate: LocalDateTime =
```
LennyDZ, line 28: You should use `clockService.now()` instead of `Instant.now().atZone(ZoneId.systemDefault())` and thus inject clockService to the class. Then you should mock a clockService for your tests.

I think using `now()`without specifying the zoneId will use the systemDefault one by default.

LennyDZ (changes requested):

LennyDZ (approved):
Looks fine to me now

PR #124, opened 2022-04-04
Display of time for the deadline

Matathiasq13824:
With the modification in the Deadline of LocalDate to LocalDateTime, it was necessary to change the display of the informations shown with the Adapter and in the DeadlineDetailsActivity. I have also added a minor change in the two of them, displaying the hours left if we are at less than a day of the deadline.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,3 +1,8 @@
+checks:
+  similar-code:
+    config:
+      threshold: 5
+
 exclude_patterns:
   - "**/test/"
   - "**/androidTest/"
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,7 +1,7 @@
 checks:
   similar-code:
     config:
-      threshold: 5
+      threshold: 250
 
 exclude_patterns:
   - "**/test/"
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,8 +1,3 @@
-checks:
-  similar-code:
-    config:
-      threshold: 250
-
 exclude_patterns:
   - "**/test/"
   - "**/androidTest/"
--- a/app/src/main/java/com/github/multimatum_team/multimatum/DeadlineDetailsActivity.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/DeadlineDetailsActivity.kt
@@ -58,17 +58,11 @@ class DeadlineDetailsActivity : AppCompatActivity() {
             }
             else -> {
                 val remainingTime = actualDate.until(date, ChronoUnit.DAYS)
-                val dueIn: Int
-                val until: String
-
-                if (remainingTime <= 0) {
-                    dueIn = R.string.DueInXHours
-                    until = actualDate.until(date, ChronoUnit.HOURS).toString()
+                detailView.text = if (remainingTime <= 0) {
+                    getString(R.string.DueInXHours, actualDate.until(date, ChronoUnit.HOURS).toString())
                 } else {
-                    dueIn = R.string.DueInXDays
-                    until = actualDate.until(date, ChronoUnit.DAYS).toString()
+                    getString(R.string.DueInXDays,actualDate.until(date, ChronoUnit.DAYS).toString())
                 }
-                detailView.text = getString(dueIn, until)
             }
         }
     }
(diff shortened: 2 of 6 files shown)


app/src/main/java/com/github/multimatum_team/multimatum/DeadlineDetailsActivity.kt
```diff
@@ -56,8 +49,18 @@ class DeadlineDetailsActivity : AppCompatActivity() {
                 detailView.text = getString(R.string.isAlreadyDue)
             }
             else -> {
-                detailView.text =
-                    getString(R.string.DueInXDays, actualDate.until(date, ChronoUnit.DAYS).toString())
+                val remainingTime = actualDate.until(date, ChronoUnit.DAYS)
+                detailView.text = if (remainingTime <= 0) {
+                    getString(
+                        R.string.DueInXHours,
+                        actualDate.until(date, ChronoUnit.HOURS).toString()
```
macintosh1398, line 56: Codeclimate fails because of code duplication. To solve that, it will be good to store R.string.DueInXHours and
actualDate.until.... in val to avoid this warning.

app/src/main/java/com/github/multimatum_team/multimatum/DeadlineDetailsActivity.kt
```diff
@@ -56,8 +49,18 @@ class DeadlineDetailsActivity : AppCompatActivity() {
...
+                    )
+                } else {
+                    getString(
+                        R.string.DueInXDays,
+                        actualDate.until(date, ChronoUnit.DAYS).toString()
```
macintosh1398, line 61: Here is the duplication

app/src/main/java/com/github/multimatum_team/multimatum/DeadlineDetailsActivity.kt
```diff
@@ -56,8 +49,18 @@ class DeadlineDetailsActivity : AppCompatActivity() {
...
+                    )
+                }
             }
         }
```
macintosh1398, line 65: The onCreate function is too long, I think some stuff could be done in private functions

macintosh1398 (changes requested):
Good work and good adaptation of the tests. Just correct the Codeclimate warnings (they are quite relevant).

app/src/main/java/com/github/multimatum_team/multimatum/DeadlineDetailsActivity.kt
Matathiasq13824: Unfortunatly, once I recuperate DueInXHours and the other time DueInXDays, it is the same for actualDate where once I recuperate the hours left and the other the days left. I can perfectly separate everything in two value that I set differently on an If but it will make the code longer and, in my opinion, will not make the code more understandable.
But if you think it is necessary, I will do it.

Matathiasq13824: I have put them in val and, infortunatly, this is still an issue for codeclimate

Matathiasq13824: Seeing that this issue is really minor but codeclimate seems to put to much importance on it, I have increased its 
threshold on similar code

app/src/main/java/com/github/multimatum_team/multimatum/DeadlineDetailsActivity.kt
```diff
@@ -43,10 +36,18 @@ class DeadlineDetailsActivity : AppCompatActivity() {
...
 
         // Set the detail text to inform the user if it is due, done or the remaining time
         val detailView = findViewById<TextView>(R.id.deadline_details_activity_done_or_due)
+        setDetailText(detailView, state, date)
+
+    }
+
+    /*
+     Set the detail text to inform the user if it is due, done or the remaining time
+     */
```
LennyDZ, line 49: It's the wrong syntax for comment, but since. Even if it's a private function, try to use "/** ... */" instead (it then appears in green in android studio)

.codeclimate.yml
```diff
@@ -1,3 +1,8 @@
+checks:
```
LennyDZ, line 1: Change implying project setup should have their own PR to avoid missunderstanding.

LennyDZ (changes requested):
Looks fine overall, but you're not using the right syntax for function comment, it should be with "/** ... */" (it appears in green in android studio)

macintosh1398 (approved):
Looks perfect now.

.codeclimate.yml
Matathiasq13824: True, I have removed it and put it in another PR

PR #128, opened 2022-04-05
Feature/add edition deadline details activity

Matathiasq13824:
This PR add the ability to edit a deadline via DeadlineDetailsActivity. It allow to change the title, the date and the state of the deadline.
Unfortunatly, I hadn't the time to link it to the view model to really modify the deadline, I will do so in the next sprint.
To have a general idea of this PR, I have added a button that will make everything editable in the activity.
The title will become a EditText, the date will become a button allowing to select a new date with a Date and TimePickerDialog and a CheckBox will appear allowing to set the deadline to done.

### DIFF
--- a/app/src/main/java/com/github/multimatum_team/multimatum/DeadlineDetailsActivity.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/DeadlineDetailsActivity.kt
@@ -1,23 +1,21 @@
 package com.github.multimatum_team.multimatum
 
+import android.app.DatePickerDialog
+import android.app.TimePickerDialog
 import android.content.Context
 import android.content.Intent
+import android.graphics.Color
 import android.os.Bundle
-import android.widget.TextView
+import android.text.SpannableStringBuilder
+import android.view.View
+import android.widget.*
 import androidx.appcompat.app.AppCompatActivity
-import com.github.multimatum_team.multimatum.model.ClockServiceEntryPoint
 import com.github.multimatum_team.multimatum.model.Deadline
 import com.github.multimatum_team.multimatum.model.DeadlineState
 import com.github.multimatum_team.multimatum.repository.DeadlineID
 import com.github.multimatum_team.multimatum.service.ClockService
-import dagger.hilt.EntryPoint
-import dagger.hilt.InstallIn
 import dagger.hilt.android.AndroidEntryPoint
-import dagger.hilt.android.EntryPointAccessors
-import dagger.hilt.components.SingletonComponent
-import java.time.LocalDate
 import java.time.LocalDateTime
-import java.time.Period
 import java.time.temporal.ChronoUnit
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/multimatum_team/multimatum/DeadlineDetailsActivity.kt
```diff
@@ -31,22 +29,131 @@ class DeadlineDetailsActivity : AppCompatActivity() {
...
+                .setImageResource(android.R.drawable.ic_menu_manage)
+
+            dateView.isClickable = false
+            dateView.setBackgroundResource(android.R.color.transparent)
+
+            doneButton.isClickable = false
+            doneButton.visibility = View.GONE
+            //TODO: modify the deadline in the repository
+
+        }
```
ValentinAebi, line 123: The `if` and the `else` branches look very similar, I think you can refactor the function to make it more concise and avoid code duplication. E.g.:
```Scala
fun goToModifyOrBack(view: View) {
   editTitle(editMode)
   findViewById<ImageButton>(R.id.deadline_details_activity_modify)
                .setImageResource(
                    if (editMode) android.R.drawable.checkbox_on_background
                    else android.R.drawable.ic_menu_manage
                )
   ...
}
```
(Extracting a function might work as well.)

app/src/main/java/com/github/multimatum_team/multimatum/DeadlineDetailsActivity.kt
```diff
@@ -31,22 +29,131 @@ class DeadlineDetailsActivity : AppCompatActivity() {
...
+    private fun editTitle(edit: Boolean) {
+        if (edit) {
+            titleView.isEnabled = true
+            titleView.setBackgroundResource(android.R.drawable.edit_text)
+            titleView.setTextColor(Color.BLACK)
+        } else {
+            titleView.isEnabled = false
+            titleView.setBackgroundResource(android.R.color.transparent)
+            titleView.setTextColor(Color.BLACK)
+        }
```
ValentinAebi, line 138: Same comment as above, even though this method is shorter so this is less of a problem

ValentinAebi (approved):

Voxanimus (commented):
I see a lot of code climate warning. Are thay too strict ?
It's a nice idea to put a done button here. I think we could also put another one in the list view in mainActivity. Create this kind of redundancy is user friendly in my opinion.

Matathiasq13824:
> I see a lot of code climate warning. Are thay too strict ? 

I think so too, it should be a good idea to increase its threshold about similar code.

> It's a nice idea to put a done button here. I think we could also put another one in the list view in mainActivity. Create this kind of redundancy is user friendly in my opinion.

Good idea ! It is a good issue for another sprint

Voxanimus (approved):

PR #129, opened 2022-04-06
Feature/generate qr code from detailed deadline

Voxanimus:
his PR made the QR generation accessible from detailed of a deadline. It takes the Firebase Id of the deadline and generate a QRCode from it.


app/src/main/java/com/github/multimatum_team/multimatum/DeadlineDetailsActivity.kt
```diff
@@ -187,6 +190,12 @@ class DeadlineDetailsActivity : AppCompatActivity() {
         timePickerDialog.show()
     }
 
+    fun goQRGenerator(view: View) {
```
Matathiasq13824, line 193: A comment here style /** ... */ to explain the function could be nice

app/src/main/java/com/github/multimatum_team/multimatum/QRGeneratorActivity.kt
```diff
@@ -17,10 +20,14 @@ import com.google.zxing.qrcode.QRCodeWriter
     to make the encoded data appear in a ImageView.
     If no data is in the TextField, a Toast will appear saying to enter some data.
  */
-class QRGenerator : AppCompatActivity() {
+class QRGeneratorActivity : AppCompatActivity() {
```
Matathiasq13824, line 23: As the class has changed, the description of it should change too

app/src/test/java/com/github/multimatum_team/multimatum/QRGeneratorActivityTest.kt
```diff
@@ -14,27 +18,35 @@ import androidx.test.espresso.matcher.BoundedMatcher
...
 
 /**
  * Tests for the QRGenerator class.
  */
 @RunWith(AndroidJUnit4::class)
-class QRGeneratorTest {
+class QRGeneratorActivityTest: QRGeneratorUtilityTest {
     @get:Rule
-    val activityRule = ActivityScenarioRule(QRGenerator::class.java)
+    val activityRule = ActivityScenarioRule(QRGeneratorActivity::class.java)
```
Matathiasq13824, line 49: As `qRDisplayTest()` need to go to QRGeneratorActivity, I'm not sure that this is a good idea to put this as it will launch the class once without any intent and I don't know how the app react to this. 
You could in a @before put 

 `val intent = Intent(ApplicationProvider.getApplicationContext(),QRGeneratorActivity::class.java).putExtra(EXTRA_ID, "1")
ActivityScenario.launch<QRGeneratorActivity>(intent)`
        
to launch all tests correctly but I'm not really sure about that. If you think that what you have done is better, do not change anything

app/src/test/java/com/github/multimatum_team/multimatum/util/QRGeneratorUtilityTest.kt
```diff
@@ -0,0 +1,35 @@
+package com.github.multimatum_team.multimatum.util
+
+import android.graphics.Bitmap
+import com.google.zxing.*
+import com.google.zxing.common.HybridBinarizer
+import com.google.zxing.qrcode.QRCodeReader
+
+interface QRGeneratorUtilityTest {
```
Matathiasq13824, line 8: Maybe add a description of the interface

Matathiasq13824 (changes requested):
The code looks fine. However they are too little comment in my opinion.

app/src/test/java/com/github/multimatum_team/multimatum/QRGeneratorActivityTest.kt
Voxanimus: what you are proposing is basically the same thing. As in my tests a sent a new intent it "update" the activity so it works too.

Matathiasq13824 (approved):

app/src/androidTest/java/com/github/multimatum_team/multimatum/AccountActivityTest.kt
```diff
@@ -5,15 +5,19 @@ import androidx.test.core.app.ActivityScenario
...
+import androidx.test.espresso.intent.matcher.IntentMatchers.hasComponent
 import androidx.test.espresso.matcher.ViewMatchers
 import androidx.test.espresso.matcher.ViewMatchers.assertThat
 import androidx.test.espresso.matcher.ViewMatchers.withId
 import androidx.test.ext.junit.rules.ActivityScenarioRule
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.filters.LargeTest
 import com.google.firebase.FirebaseApp
+import com.google.firebase.auth.FirebaseAuth
+import org.hamcrest.Matchers.allOf
```
felko, line 20: Please remove unused imports here

app/src/test/java/com/github/multimatum_team/multimatum/util/QRGeneratorUtilityTest.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import android.graphics.Bitmap
+import com.google.zxing.*
+import com.google.zxing.common.HybridBinarizer
+import com.google.zxing.qrcode.QRCodeReader
+
+/**
+ * Interface that contain utility methods use in QRGeneratorTest and QRGeneratorActivityTest
+ */
+interface QRGeneratorUtilityTest {
```
felko, line 11: Instead of defining an interface where all methods are already implemented, you could just write two functions at the toplevel.

felko (changes requested):

app/src/test/java/com/github/multimatum_team/multimatum/util/QRGeneratorUtilityTest.kt
Voxanimus: Well that's right. At begining i thought it would be a good idea to make an interface because these function are used by only 2 classe.

felko (approved):

PR #136, opened 2022-04-06
Feature/procrastination detector foreground service

### LINKED ISSUE #102: Make the procrastination detector service run when the app is closed

ValentinAebi:
Procrastination detector now also works when the app is swiped out. It uses a foreground service for that, so a notification is displayed as long as the service is active.

This feature is still experimental, the detection is currently too sensitive to be usable, but this will be adjusted in future sprint(s). In particular we will need to let the user access Multimatum without being harassed by injunctions to start working...

Closes #102

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -36,7 +36,7 @@
             android:name=".QRCodeReaderActivity"
             android:exported="false" />
         <activity android:name=".MainSettingsActivity"
-            android:exported="true">
+            android:exported="false">
         </activity>
         <activity
             android:name=".AccountActivity"
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -5,6 +5,8 @@
 
     <uses-permission android:name="android.permission.SCHEDULE_EXACT_ALARM" />
     <uses-permission android:name="android.permission.CAMERA" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
+    <uses-permission android:name="android.permission.WAKE_LOCK" />
 
     <application
         android:name=".MultimatumApp"
@@ -33,9 +35,9 @@
         <activity
             android:name=".QRCodeReaderActivity"
             android:exported="false" />
-        <activity
-            android:name=".MainSettingsActivity"
-            android:exported="false" />
+        <activity android:name=".MainSettingsActivity"
+            android:exported="true">
+        </activity>
         <activity
             android:name=".AccountActivity"
             android:exported="false"
@@ -49,7 +51,6 @@
             android:exported="true">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
-
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
--- a/app/src/main/java/com/github/multimatum_team/multimatum/service/ProcrastinationDetectorService.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/service/ProcrastinationDetectorService.kt
@@ -1,14 +1,19 @@
 package com.github.multimatum_team.multimatum.service
 
-import android.app.Service
+import android.app.*
+import android.content.Context
 import android.content.Intent
+import android.graphics.Color
 import android.hardware.Sensor
 import android.hardware.SensorEvent
 import android.hardware.SensorEventListener
 import android.hardware.SensorManager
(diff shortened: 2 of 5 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -33,9 +35,9 @@
         <activity
             android:name=".QRCodeReaderActivity"
             android:exported="false" />
-        <activity
-            android:name=".MainSettingsActivity"
-            android:exported="false" />
+        <activity android:name=".MainSettingsActivity"
+            android:exported="true">
```
Voxanimus, line 39: what does exported put at true ?

app/src/main/java/com/github/multimatum_team/multimatum/service/ProcrastinationDetectorService.kt
```diff
@@ -93,14 +192,48 @@ class ProcrastinationDetectorService : Service(), SensorEventListener {
          * The sensor used by this service to detect movements
          */
         const val REF_SENSOR = Sensor.TYPE_GRAVITY
-      
-        private const val MIN_PERIOD_BETWEEN_NOTIF_NANOSEC = 2_000_000_000L
+
+        private const val MIN_PERIOD_BETWEEN_NOTIF_NANOSEC = 2_000_000_000L  // 2 seconds
```
Voxanimus, line 197: ahy this is noted as a change ?

app/src/main/res/values/strings.xml
```diff
@@ -29,6 +29,8 @@
     <string name="logout">LOGOUT</string>
     <string name="procrastination_fighter_enable_msg">Procrastination fighter is enabled</string>
     <string name="procrastination_fighter_disabled_msg">Procrastination fighter is disabled</string>
+    <string name="procrastination_fighter_permanent_notif_title">Procrastination detector</string>
+    <string name="procrastination_fighter_permanent_notif_content_text">Procrastination is bad...</string>
     <string name="deleted">Deleted</string>
     <string name="undo">Undo</string>
     <string name="write_title_here">Write Title here</string>
```
Voxanimus, line 36: think too clean unuse string, and pay attention to Cirrus warning

Voxanimus (changes requested):

app/src/main/java/com/github/multimatum_team/multimatum/service/ProcrastinationDetectorService.kt
ValentinAebi: Because of the comment that I added at the end of the line

app/src/main/AndroidManifest.xml
ValentinAebi: I think that is a mistake, thanks

app/src/main/res/values/strings.xml
Voxanimus: It seems that cirrus warning disapear. So all is good

Voxanimus (approved):

Matathiasq13824 (changes requested):
Fine code, however you need to add multiple comment to describe multiple functions (mainly the private ones `ProcrastinationDetectorService.kt` and `ProcrastinationDetectorServiceTest.kt`. If someone else have to modify your code, it has to understand what you have done and the comments are the best for that.

Except this, everything is fine. The total coverage has decreased but, based on your diff-coverage which is really good, it seems that codeclimate only compare the coverage of the actual main without thinking that we had maybe increased its coverage with the merge of another branch between the creation of this branch and now. You can merge the main on this branch to resolve this (like when we resolve conflict) or just let it like that.

app/src/main/java/com/github/multimatum_team/multimatum/service/ProcrastinationDetectorService.kt
```diff
@@ -17,38 +22,131 @@ import kotlin.math.abs
...
     private var lastDetectionTimestamp: Long = 0L
     private var lastPosition: Array<Float>? = null  // null only at initialization
 
     private val binder = PdsBinder()
 
+    override fun onBind(intent: Intent): IBinder {
+        return binder
+    }
+
+    // also used to stop the service, when action is STOP_ACTION
```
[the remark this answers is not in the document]
ValentinAebi: I agree that this comment is unclear and I will change it. However, I tried to avoid the need for comments by making the code modular. As you can see, functions are short and their names describe what they do, instead of comments. So I do not really understand what additional comments I should write...

Matathiasq13824 (approved):

PR #138, opened 2022-04-07
Add notification when creating deadline

LennyDZ:
In this branch I add 2 attribute to deadline; description and notificationTime.

I also add some UI to add notification when creating a new deadline, this comes with some util functions to set/cancel all notification for a deadline.

This is not functional, I still need to call set/cancel notification when a deadline is added/removed from Firebase. I'll have to create listener for this.

### DIFF
--- a/app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
@@ -43,6 +43,14 @@ class AddDeadlineActivity : AppCompatActivity() {
         setContentView(R.layout.activity_add_deadline)
         textDate = findViewById(R.id.add_deadline_text_date)
         textTime = findViewById(R.id.add_deadline_text_time)
+        findViewById<CheckBox>(R.id.radio_notification_1h).text =
+            getString(R.string.notify_before, "1 hour")
+        findViewById<CheckBox>(R.id.radio_notification_5h).text =
+            getString(R.string.notify_before, "5 hours")
+        findViewById<CheckBox>(R.id.radio_notification_1d).text =
+            getString(R.string.notify_before, "1 day")
+        findViewById<CheckBox>(R.id.radio_notification_3d).text =
+            getString(R.string.notify_before, "3 days")
         selectedDate = clockService.now().truncatedTo(ChronoUnit.HOURS)
         textDate.text = selectedDate.toLocalDate().toString()
         textTime.text = selectedDate.toLocalTime().toString()
--- a/app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
@@ -59,14 +59,17 @@ class AddDeadlineActivity : AppCompatActivity() {
                     // monthOfYear is based on an enum that begin with 0
                     // thus we need to increase it by one to have the true date
                     year, monthOfYear + 1, dayOfMonth,
-                    selectedDate.hour, selectedDate.minute)
+                    selectedDate.hour, selectedDate.minute
+                )
 
                 textDate.text = selectedDate.toLocalDate().toString()
             }
 
         // Prepare the datePickerDialog
-        val datePickerDialog = DatePickerDialog(this, dateSetListener,
-            selectedDate.year, selectedDate.month.ordinal, selectedDate.dayOfMonth)
+        val datePickerDialog = DatePickerDialog(
+            this, dateSetListener,
+            selectedDate.year, selectedDate.month.ordinal, selectedDate.dayOfMonth
+        )
 
         // Show the Dialog on the screen
         datePickerDialog.show()
@@ -81,29 +84,34 @@ class AddDeadlineActivity : AppCompatActivity() {
             TimePickerDialog.OnTimeSetListener { _, hoursOfDay, minutes ->
                 selectedDate = LocalDateTime.of(
                     selectedDate.year, selectedDate.monthValue, selectedDate.dayOfMonth,
-                    hoursOfDay, minutes)
+                    hoursOfDay, minutes
+                )
 
                 textTime.text = selectedDate.toLocalTime().toString()
             }
 
         // Prepare the datePickerDialog
-        val timePickerDialog = TimePickerDialog(this, timeSetListener,
-            selectedDate.hour, selectedDate.minute, true)
+        val timePickerDialog = TimePickerDialog(
+            this, timeSetListener,
+            selectedDate.hour, selectedDate.minute, true
+        )
 
         // Show the Dialog on the screen
         timePickerDialog.show()
 
     }
+
     val checkBoxIdTime = mapOf<Int, Long>(
         R.id.radio_notification_1h to Duration.ofHours(1).toMillis(),
         R.id.radio_notification_5h to Duration.ofHours(5).toMillis(),
         R.id.radio_notification_1d to Duration.ofDays(1).toMillis(),
         R.id.radio_notification_3d to Duration.ofDays(3).toMillis()
     )
+
     private fun retrieveNotificationsTimes(): ArrayList<Long> {
         var res = ArrayList<Long>()
-        for (x in checkBoxIdTime){
-            if(findViewById<CheckBox>(x.key).isChecked) res.add(x.value)
+        for (x in checkBoxIdTime) {
+            if (findViewById<CheckBox>(x.key).isChecked) res.add(x.value)
         }
         return res
     }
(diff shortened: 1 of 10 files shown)


felko (changes requested):
Besides the comment on string resources, everything looks fine to me.
Some files could be reformatted.

felko (approved):

Voxanimus (approved):

app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
```diff
@@ -78,20 +92,38 @@ class AddDeadlineActivity : AppCompatActivity() {
...
         timePickerDialog.show()
 
     }
 
+    val checkBoxIdTime = mapOf<Int, Long>(
+        R.id.radio_notification_1h to Duration.ofHours(1).toMillis(),
+        R.id.radio_notification_5h to Duration.ofHours(5).toMillis(),
+        R.id.radio_notification_1d to Duration.ofDays(1).toMillis(),
+        R.id.radio_notification_3d to Duration.ofDays(3).toMillis()
+    )
```
ValentinAebi, line 117: That is a suggestion and I am not sure whether it is feasible, but as you have many different data relative to these reminders, you should think of creating a data class, something like
```Kotlin
data class Reminder(delayMillis: Long, delayText: String, resourceId: Int)
...
val reminder1Hour = Reminder(Duration.ofHours(1).toMillis(), "1 hour", R.id.radio_notification_1h)
```
and put all the reminders into a single list. This would be less error-prone and make it easier to add reminders. And it would suppress the need for the constants that I said in my previous comment.

app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
```diff
@@ -78,20 +92,38 @@ class AddDeadlineActivity : AppCompatActivity() {
...
+
+    private fun retrieveNotificationsTimes(): ArrayList<Long> {
```
ValentinAebi, line 119: Nitpicking, but do you have a good reason to use `ArrayList` instead of `List` as the return type?

app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
```diff
@@ -78,20 +92,38 @@ class AddDeadlineActivity : AppCompatActivity() {
...
+        var res = ArrayList<Long>()
+        for (x in checkBoxIdTime) {
+            if (findViewById<CheckBox>(x.key).isChecked) res.add(x.value)
+        }
+        return res
```
ValentinAebi, line 124: This could probably be done functionally, with something like:
```Kotlin
return checkBoxIdTime
  .filter{ findViewById<CheckBox>(x.key).isChecked }
  .map{ it.value }
```
Otherwise you can keep it imperative, but in this case `res` should be a `val` and you should use an identifier that is more explicit than `x`.

app/src/main/java/com/github/multimatum_team/multimatum/DeadlineNotification.kt
```diff
@@ -80,23 +80,32 @@ class DeadlineNotification {
 
     }
 
+    /**
+     * Set all the notification of a deadline's notificationsTimes list.
+     */
+    fun setDeadlineNotifications(deadlineId: DeadlineID, deadline: Deadline, context: Context) {
+        for (x in deadline.notificationsTimes) {
+            setNotification(deadlineId + x.toString(), deadline, context, x)
```
ValentinAebi, line 88: Code would be more readable using a more explicit identifier than `x`

app/src/main/java/com/github/multimatum_team/multimatum/DeadlineNotification.kt
```diff
@@ -105,4 +114,13 @@ class DeadlineNotification {
             context.getSystemService(AppCompatActivity.ALARM_SERVICE) as AlarmManager  //this get an service instance of AlarmManager
         alarmManager.cancel(pendingIntent)
     }
+
+    /**
+     * delete all notification of a deadline's notificationsTimes list.
+     */
+    fun cancelDeadlineNotifications(deadlineId: DeadlineID, deadline: Deadline, context: Context) {
+        for (x in deadline.notificationsTimes) {
+            cancelNotification(deadlineId + x.toString(), deadline, context)
```
ValentinAebi, line 123: Same comment as before, about `x`

app/src/main/java/com/github/multimatum_team/multimatum/model/Deadline.kt
```diff
@@ -30,14 +30,16 @@ enum class DeadlineState {
...
  * @constructor Creates a deadline from specified parameters
  * @throws IllegalArgumentException when title is empty or startDate > end
  */
 data class Deadline(
     val title: String,
     val state: DeadlineState,
     val dateTime: LocalDateTime,
-    val description: String = ""
+    val description: String = "",
+    val notificationsTimes: ArrayList<Long> = ArrayList()
```
ValentinAebi, line 42: As before, unless I missed something type should be `List<Long>` (or maybe `MutableList` if you want to make sure that it is mutable)

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseDeadlineRepository.kt
```diff
@@ -48,7 +50,22 @@ class FirebaseDeadlineRepository @Inject constructor() : DeadlineRepository {
             .ofEpochMilli(milliseconds)
             .atZone(ZoneId.systemDefault())
             .toLocalDateTime()
-        return Deadline(title, state, date)
+        var description = ""
+        if (deadlineSnapshot["description"] != null) {
+            description = deadlineSnapshot["description"] as String
+        }
```
ValentinAebi, line 56: I think it could be simplified:
```Kotlin
val snapshot = deadlineSnapshot["description"]
val description = if (snapshot != null) (snapshot as String) else ""
```
And Android Studio will possibly advise something better using the Elvis operator, or something similar.

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseDeadlineRepository.kt
```diff
@@ -48,7 +50,22 @@ class FirebaseDeadlineRepository @Inject constructor() : DeadlineRepository {
...
+        var notificationsTimes = ArrayList<Long>()
+        if (deadlineSnapshot["notificationsTimes"] != null) {
+            notificationsTimes = deadlineSnapshot["notificationsTimes"] as ArrayList<Long>
+        }
```
ValentinAebi, line 60: The same kind of simplification probably applies here too.

app/src/test/java/com/github/multimatum_team/multimatum/AddDeadlineTest.kt
```diff
@@ -91,16 +102,38 @@ class AddDeadlineTest {
...
 
+        onView(withId(R.id.radio_notification_1h)).perform()
+
         // Check if Toast correctly appear
         onView(withId(R.id.add_deadline_button)).perform(ViewActions.click())
-        MatcherAssert.assertThat(ShadowToast.getTextOfLatestToast(),
-            CoreMatchers.equalTo(RuntimeEnvironment.getApplication().applicationContext.getString(R.string.deadline_created)))
+        MatcherAssert.assertThat(
+            ShadowToast.getTextOfLatestToast(),
+            CoreMatchers.equalTo(RuntimeEnvironment.getApplication().applicationContext.getString(R.string.deadline_created))
```
ValentinAebi, line 117: You can get the `applicationContext` once and at the top of the class and simply use a variable for it, it would be more readable

app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
LennyDZ: well I though we might like to add/remove notification over time.

app/src/main/java/com/github/multimatum_team/multimatum/AddDeadlineActivity.kt
LennyDZ: ok

app/src/main/java/com/github/multimatum_team/multimatum/model/Deadline.kt
LennyDZ: I think you're right

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseDeadlineRepository.kt
LennyDZ: ok

LennyDZ: ça fait crash le programme si je fait ça donc je vais laisser comme avant

Issue #125, opened 2022-04-05
Link authentication activity so that new deadlines are associated to the signed-in account


felko:
Since we already have the login/signin activity, we should finally make use of it by associating deadlines with the ID of the user that created it (see #83).

felko:
Closed by #139

PR #139, opened 2022-04-07
Implement users

### LINKED ISSUE #83: Associate deadlines to signed-in account
Currently Firebase stores a single global collection of deadlines shared by all instances of the application, regardless of the signed-in account, if any. We should integrate the Firebase authentication to the deadline repository so that each user has their own list of deadlines, and cannot access the deadlines of other users.

felko:
This PR implements users to the application. Deadlines are now associated to an user ID.
A non authenticated user can still use the application, thanks to the anonymous users feature of Firebase.
At the first start of the application, an anonymous user is created. When the user signs in with a google account, all deadlines are associated to the new user.
When logging out, a new anonymous user is created. This lets us assume in the entire codebase that thee's always an user to which we can associate the deadlines.

Closes #83, #125


app/src/androidTest/java/com/github/multimatum_team/multimatum/MainActivityTest.kt
```diff
@@ -237,5 +240,14 @@ class MainActivityTest {
...
                 )
             )
+
+        @Singleton
+        @Provides
+        fun provideAuthRepository(): AuthRepository =
+            MockAuthRepository()
+    }
+
+    companion object {
```
Matathiasq13824, line 250: I'm not really sure. Why this need to be in a companion object ?

app/src/main/java/com/github/multimatum_team/multimatum/ProfileActivity.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.github.multimatum_team.multimatum.model.User
+import com.github.multimatum_team.multimatum.viewmodel.UserViewModel
+import dagger.hilt.android.AndroidEntryPoint
+
+/**
+ * A simple [Fragment] subclass.
+ * Use the [ProfileActivity.newInstance] factory method to
+ * create an instance of this fragment.
+ */
+@AndroidEntryPoint
```
Matathiasq13824, line 19: You could change this comment to explain briefly what this fragment do

app/src/main/java/com/github/multimatum_team/multimatum/SignInActivity.kt
```diff
@@ -0,0 +1,73 @@
...
+import android.widget.Toast
+import androidx.appcompat.app.AppCompatActivity
+import com.firebase.ui.auth.AuthUI
+import com.firebase.ui.auth.FirebaseAuthUIActivityResultContract
+import com.firebase.ui.auth.data.model.FirebaseAuthUIAuthenticationResult
+import com.google.firebase.auth.FirebaseAuth
+import dagger.hilt.android.AndroidEntryPoint
+
+@AndroidEntryPoint
+class SignInActivity : AppCompatActivity() {
```
Matathiasq13824, line 16: You could add comments to describe the class and the functions

Matathiasq13824 (changes requested):
Great job, it was a collosal work ! However please next time try to separate it on multiple pr (35 files is quite a lot ^^')
I have only put some minor comments. However, it could be nice to add some tests to increase coverage in the next sprint. When the CI has finished, the test pass and some minor modification are made, I will approve

app/src/androidTest/java/com/github/multimatum_team/multimatum/MainActivityTest.kt
felko: Before I moved this test to the roboelectric tests, the mockSharedPreferences was taken from the MainSettingsActivityTest, which I kept in the instrumented tests. Therefore I simply copied what was in MainSettingsActivityTest and put it in MainActivityTest.

macintosh1398 (approved):
What a great work, a lot of new code but the modularity is good and the code is clean. The previous comment is relevant, so nothing to add. Maybe just focus on the tests next week.

Matathiasq13824 (approved):

PR #144, opened 2022-04-08
Fix UI in main

Voxanimus:
Little fix after demo.
Remove QRGenerator button in main, it's not suppose to be here.


macintosh1398 (approved):
Thank you for the correction

Matathiasq13824 (approved):

PR #152, opened 2022-04-13
Tuning/procrastination detector sensitivity

### LINKED ISSUE #140: Make procrastination detector less sensitive

### LINKED ISSUE #153: Make a module with utility logging methods

ValentinAebi:
I modified the procrastination detector so that it is more handy to use. Now, the procrastination detector service stops working (and sending toasts) when the app comes to the foreground. It is also a bit less sensitive and after it was started it lets a few seconds to the user to lay down her/his phone before displaying a toast.

This PR also includes `LogUtil`, a module that I designed to make logging easier because I had to use logging extensively to debug the code and understand the Android activity lifecycle.

<del> Finally, I moved the constants containing the keys of the SharedPreferences from `MainSettingsActivity` to a separate class, `SharedPreferenceKeys`. This may cause issues when merging if someone used the old version of these constants, we have to keep aware of that.</del> Constants are back in `MainSettingsActivity` to try to avoid issues.

Closes #140 
Closes #153


ValentinAebi:
Fixing the issue found by CodeClimate (too many functions in the same file) will require to move some methods to another file. This will probably take some time as the tests may need to be adapted, so I will do it during another sprint.

app/src/main/java/com/github/multimatum_team/multimatum/MultimatumApp.kt
```diff
@@ -1,13 +1,109 @@
...
+    }
+
+    override fun onActivityPaused(activity: Activity) {
+        logFunctionCall("paused ${activity.localClassName}")
+    }
+
+    override fun onActivityStopped(activity: Activity) {
+        logFunctionCall("stopped ${activity.localClassName}")
+        val onForeground = isAppOnForeground()
+        debugLog(if (onForeground) "app is on foreground" else "app is on background")
```
macintosh1398, line 52: Are the debugLog still useful ? If they are, you can ignore my comment

macintosh1398 (approved):
Good job Valentin! You have a (very) modular code, every number are well defined as constants and your tests provide a great coverage, but are very relevant as well. I just two things to add: you have a lot of logs, if they are useful you can keep them of course, but sometimes we only use them to debug and they lose their purpose after that, so think about it. And has you said yourself, you have a lot of functions, and sometimes too much modularity in one file can be bad for the maintainability, but maybe you can ask the coach about that. But you did a great job!

app/src/main/java/com/github/multimatum_team/multimatum/MultimatumApp.kt
ValentinAebi: I think the logs are not a problem for now, and may be useful when we debug the app in the following weeks. But yes at the very end of the project the ideal thing to do will probably be to remove them.

felko: We could probably setup the LogUtils so that the logs are only emitted on debug mode, and ignored in release mode.

app/src/main/java/com/github/multimatum_team/multimatum/MultimatumApp.kt
```diff
@@ -1,13 +1,109 @@
...
+    override fun onActivitySaveInstanceState(activity: Activity, outState: Bundle) {
+        logFunctionCall()
+    }
+
+    override fun onActivityDestroyed(activity: Activity) {
+        logFunctionCall("destroyed ${activity.localClassName}")
+    }
+
+    @Inject
+    lateinit var sharedPreferences: SharedPreferences
```
felko, line 67: Nitpick: I think it would be preferable to use the convention that all the attribute definitions (including injects) appear before the method definitions.

felko (approved):
All good! The LogUtils class will come in handy for debugging.

app/src/main/java/com/github/multimatum_team/multimatum/MultimatumApp.kt
ValentinAebi: Indeed, and I think that this is a bit more important than nitpicking!

ValentinAebi: > We could probably setup the LogUtils so that the logs are only emitted on debug mode, and ignored in release mode.

I thought of it too actually :) . But how would you define debug mode? Using a constant or do you see a better solution?

PR #154, opened 2022-04-14
Link DeadlineDetailsActivity to the view model

### LINKED ISSUE #131: Link the modification of the deadline to the viewModel

Matathiasq13824:
Previously, the `DeadlineDetailsActivity` had the option to modify the information shown but this didn't changed the actual deadline in the database. This is now solved by adding the viewModel in it. 
As we have now the view model where we can recuperate the informations of the deadline, we only need to put, in the intent to launch the activity, the id of the chosen deadline. However, the viewModel can take time to fetch the data from the database so we can't recuperate it immediatly on the `onCreate()` so we put default value first (More precisely "") before updating them when the deadline are fetched.
Aditionnaly, this pr clean a little the code of `AddDeadlineActivity` and `AddDeadlineActivityTest`, removing some warning.
This pr close #131


app/src/main/java/com/github/multimatum_team/multimatum/DeadlineDetailsActivity.kt
```diff
@@ -32,40 +34,52 @@ class DeadlineDetailsActivity : AppCompatActivity() {
...
-        dateView.text = getString(R.string.DueTheXatX, date.toLocalDate(), date.toLocalTime())
+        // Recuperate the id of the deadline
+        id = intent.getStringExtra(EXTRA_ID) as DeadlineID
+
+        // As the viewModel doesn't recuperate immediately the deadlines,
+        // we need an update the moment they are fetched
+        deadlineListViewModel.getDeadlines().observe(this) { deadlines ->
+            // Recuperate the data from the deadline
+            val deadline = deadlines[id]!!
+            date = deadline.dateTime
```
ValentinAebi, line 67: Is this a `Date` or a `DateTime`? I suggest you rename the variable for clarity.

app/src/main/java/com/github/multimatum_team/multimatum/viewmodel/DeadlineListViewModel.kt
```diff
@@ -59,6 +60,14 @@ class DeadlineListViewModel @Inject constructor(
     fun getDeadlines(): LiveData<Map<DeadlineID, Deadline>> =
         _deadlines
 
+    /**
+     * get a deadline from the repository.
+     */
+    fun getDeadline(id: DeadlineID): Deadline =
+        _deadlines.value!![id]!!
```
ValentinAebi, line 67: Sanity check, but there seems to be a lot of `!!` here. Are you sure that nothing can be null?

ValentinAebi, line 67: Sanity check, but there seems to be a lot of `!!` here. Are you sure that nothing can be null here?

app/src/test/java/com/github/multimatum_team/multimatum/DeadlineDetailsTest.kt
```diff
@@ -57,9 +61,8 @@ class DeadlineDetailsTest {
 
     @Test
     fun `Given a deadline not yet due or done, the activity should display it`() {
-        val intent = DeadlineDetailsActivity.newIntent(ApplicationProvider.getApplicationContext(),
-            "1", Deadline("Test 1", DeadlineState.TODO, clockService.now().plusDays(7))
-        )
+        val intent = DeadlineDetailsActivity.newIntent(ApplicationProvider.getApplicationContext(), "0" )
```
ValentinAebi, line 64: What is this `"0"` argument?

ValentinAebi, line 64: What is this `"0"` argument?

app/src/main/java/com/github/multimatum_team/multimatum/viewmodel/DeadlineListViewModel.kt
Matathiasq13824: True, I have put this to make some tests but with the observe that I have now, this can be removed

app/src/test/java/com/github/multimatum_team/multimatum/DeadlineDetailsTest.kt
Matathiasq13824: Now, the intent send only the id of the deadline, which is the position in the list of MockDeadlineRepository at the end of the test class. As we want to check the firt deadline of the list, we put 0 (like every list in coding, we begin a list at 0)

app/src/main/java/com/github/multimatum_team/multimatum/DeadlineDetailsActivity.kt
```diff
@@ -32,40 +34,52 @@ class DeadlineDetailsActivity : AppCompatActivity() {
...
-        id = intent.getStringExtra(EXTRA_ID) as DeadlineID
-        date = intent.getSerializableExtra(EXTRA_DATE) as LocalDateTime
-        state = intent.getSerializableExtra(EXTRA_STATE) as DeadlineState
+        var title: String
 
         // Recuperate the necessary TextView
         titleView = findViewById(R.id.deadline_details_activity_title)
         dateView = findViewById(R.id.deadline_details_activity_date)
         detailView = findViewById(R.id.deadline_details_activity_done_or_due)
         doneButton = findViewById(R.id.deadline_details_activity_set_done)
```
macintosh1398, line 57: As said by CodeClimate, this function should be refactored. The variables initialization could be done outside easily, as well as the DeadlineListViewModel observation below, and the shown data update (end of function).

macintosh1398 (commented):
Your code is good. As usual, well commented and the modularity is good. @ValentinAebi made relevant comments/questions. Personally I have nothing to add here outside my code comment. I have seen that you have trouble with the QR-Code reader test (which is mine actually). It was made at the very beginning of the project with the same structure as many other tests. It is weitd because I never had problems with it. It relies on Intents, so many there could be a new intent conflict that appears with new tests ?

app/src/main/java/com/github/multimatum_team/multimatum/DeadlineDetailsActivity.kt
```diff
@@ -32,40 +34,38 @@ class DeadlineDetailsActivity : AppCompatActivity() {
...
     private lateinit var titleView: EditText
     private lateinit var dateView: TextView
     private lateinit var detailView: TextView
     private lateinit var doneButton: CheckBox
 
-    private lateinit var date: LocalDateTime
-    private lateinit var state: DeadlineState
+    // Set them on default value, waiting the fetch of the deadlines
+    private var dateTime: LocalDateTime = LocalDateTime.of(2022, 10, 10, 10, 10)
+    private var state: DeadlineState = DeadlineState.TODO
```
LennyDZ, line 46: Do you really need this? Can't you just leave lateinit?

LennyDZ (approved):
Seems fine to me, just ensure to check the little comment I made.

app/src/main/java/com/github/multimatum_team/multimatum/DeadlineDetailsActivity.kt
LennyDZ: And how about to use a null value to avoid confusion?

Matathiasq13824: It is not possible as `LocalDateTime` and `DeadlineState` are non-null type and thus can't have a null value

ValentinAebi (approved):

PR #162, opened 2022-04-23
notification locally stored

LennyDZ:
With the rise of users features and shared deadline, it seems dangerous to have notification that are shared through users. A malicious user could set notification to someone else's phone which is bad.

So now notifications are locally stored and manage from a local sharedPreferences file.

The file `DeadlineNotification` has been moved to an `util` package.


app/src/main/java/com/github/multimatum_team/multimatum/util/DeadlineNotification.kt
```diff
@@ -0,0 +1,258 @@
...
+        val notifications = sp.all.keys //key of deadline that have notification
+        val existingDeadline = deadlineList.keys //keys of existing deadline
+        val editor: SharedPreferences.Editor = sp.edit()
+        for(k in notifications) {
+            if (!existingDeadline.contains(k)){ //if deadlineId isn't in the list of deadline, then remove the notifications
+                editor.remove(k)
+                cancelDeadlineNotifications(k)
+            }
+        }
+        editor.commit()
```
ValentinAebi, line 88: Cirrus suggests you use `apply` instead of `commit`, and I think that usually it is right, so think of it...

app/src/main/java/com/github/multimatum_team/multimatum/util/DeadlineNotification.kt
```diff
@@ -0,0 +1,258 @@
...
+
+    /**
+     * Iterate through existing notification to ensure that there isn't any notification that aren't bound to a deadline.
+     * If such a notification is detected, it'll delete it.
+     */
+    fun updateNotifications(deadlineList:Map<DeadlineID, Deadline>) {
+        val sp = context.getSharedPreferences(
+            context.getString(R.string.notification_shared_preference),
+            Context.MODE_PRIVATE
+        )
```
ValentinAebi, line 78: Why don't you use the same `SharedPreferences` instance as the rest of the app? It would be easier and it would allow you to use dependency injection to test this class.

app/src/main/java/com/github/multimatum_team/multimatum/util/DeadlineNotification.kt
```diff
@@ -0,0 +1,258 @@
...
+     */
+    private fun writeToSharedPreference(id: String, value: List<Long>) {
+        val jsonString = Gson().toJson(value)
+        val sharedPref: SharedPreferences = context.getSharedPreferences(
+            context.getString(R.string.notification_shared_preference),
+            Context.MODE_PRIVATE
+        )
+        var editor = sharedPref.edit()
+        editor.putString(id, jsonString)
+        editor.commit()
```
ValentinAebi, line 103: `editor` could be a `val`. And as above `apply` might be better than `commit`.

app/src/main/java/com/github/multimatum_team/multimatum/util/DeadlineNotification.kt
```diff
@@ -0,0 +1,258 @@
...
+            ReminderBroadcastReceiver::class.java
+        )
+        //this create an intent of broadcast receiver
+        //Adding extra parameter that will be used in the broadcast receiver to create the notification
+        intent.putExtra("title", deadline.title)
+        intent.putExtra("description", deadline.description)
+        intent.putExtra("id", id)
+
+        //compute the time where the alarm will be triggered in millis.
+        val alarmTriggerTime:Long = deadline.dateTime.atZone(ZoneId.systemDefault()).toInstant()
```
ValentinAebi, line 185: Maybe specify the time unit in the variable name to avoid issues.

app/src/main/java/com/github/multimatum_team/multimatum/util/DeadlineNotification.kt
```diff
@@ -0,0 +1,258 @@
...
+                    context,
+                    id.hashCode(),
+                    intent,
+                    PendingIntent.FLAG_IMMUTABLE
+                )
+
+            //set an alarm that will wake up the pending intent (receiver)
+            alarmManager.setExact(AlarmManager.RTC_WAKEUP, alarmTriggerTime, pendingIntent)
+        } else {
+            Log.i("not sheduled", "alarm Already Passed")
```
ValentinAebi, line 200: Typo in log message: "scheduled"

app/src/main/java/com/github/multimatum_team/multimatum/viewmodel/DeadlineListViewModel.kt
```diff
@@ -36,10 +39,15 @@ class DeadlineListViewModel @Inject constructor(
...
         }
 
+        deadlineNotification = DeadlineNotification(getApplication<Application>().applicationContext)
+        _deadlines.value?.let { deadlineNotification.updateNotifications(it) }
+
         // Listen for authentication updates, upon which the deadline list is re-fetched
         authRepository.onUpdate {
             deadlineRepository.setUser(it)
-            refreshDeadlines()
+            refreshDeadlines() { it2 ->
```
ValentinAebi, line 48: A more descriptive name than `it2` would probably make the code more readable.

app/src/test/java/com/github/multimatum_team/multimatum/DeadlineNotificationTest.kt
```diff
@@ -78,126 +80,70 @@ class DeadlineNotificationTest {
...
-        deadlineNotification.setDeadlineNotifications(id, deadline, context)
-        Assert.assertEquals(2, shadowAlarmManager.scheduledAlarms.size)
-
-        //delete notifications and verify that alarmManager are indeed empty
-        deadlineNotification.cancelDeadlineNotifications(id, deadline, context)
-        Assert.assertEquals(0, shadowAlarmManager.scheduledAlarms.size)
+    private fun getNotificationsSample(tag: Long): List<Long>{
+        val notif1: Long = 1000 + tag
+        val notif2: Long = Duration.ofDays(1).toMillis()
+        return arrayListOf<Long>(notif1, notif2)
```
ValentinAebi, line 146: Is there a particular reason to use `arrayListOf` instead of simply `listOf`?

ValentinAebi (changes requested):

app/src/main/java/com/github/multimatum_team/multimatum/util/DeadlineNotification.kt
```diff
@@ -0,0 +1,258 @@
...
+import javax.inject.Inject
+
+
+/**
+ * This service provides function to manage notifications
+ * It provide functions to access and modifiy notifications that are set for a deadline,
+ * A function that verify that notification are bound to existing deadline,
+ * and a function to create notification channel
+ */
+class DeadlineNotification(context: Context) {
```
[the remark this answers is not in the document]
LennyDZ: Sure, I don't use it. I just forgot to remove it

app/src/main/java/com/github/multimatum_team/multimatum/util/DeadlineNotification.kt
```diff
@@ -0,0 +1,258 @@
...
+     * return a list of all notifications set for this deadline
+     */
+    fun listDeadlineNotification(deadlineId: DeadlineID): List<Long> {
+        val res = readFromSharedPreference(deadlineId)
+        //this is just to remove decimal to the number
+        var resParsed = ArrayList<Long>()
+        for (x in res) {
+             resParsed.add(x.toLong())
+        }
+        return resParsed
```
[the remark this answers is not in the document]
LennyDZ: I'm fine with functionnal style, I'll apply the change you suggested

app/src/main/java/com/github/multimatum_team/multimatum/util/DeadlineNotification.kt
LennyDZ: Well I think it is more clear to have notification in another file, since there may be many of them. And furthermore to retrieve all notification, we can just call getAll() on this file, so we don't have to filter other value

app/src/test/java/com/github/multimatum_team/multimatum/DeadlineNotificationTest.kt
LennyDZ: No, I changed that

app/src/main/java/com/github/multimatum_team/multimatum/util/DeadlineNotification.kt
LennyDZ: I changed that too, I was afraid of data race because of apply being async, but after further investigation, it seems that it is well handled by the implementation.

ValentinAebi (approved):

Voxanimus (approved):
For me it's a good improvement. And all the modification more than just improve security seems also to be more readable.

PR #167, opened 2022-04-23
Add tests for the viewmodel

### LINKED ISSUE #142: Add tests for new functions introduced by the users feature
The users PR (#139) introduced a few testable functions that I did not have the time to test, resulting in a slight drop in the coverage.
We should add tests for those.

felko:
This PR tests a couple of functions in the DeadlineListViewModel, namely `getDeadline`, `addDeadline`, `deleteDeadline`, and `modifyDeadline`.

Closes #142 even though I am still looking for a proper way to test the actual firebase code using the firebase emulator suite.


Voxanimus (commented):
every things seems good. And just to be sure. The callback you use in DeadlineListViewModel are here to works asynchronously  with furebase that's right ?

app/src/main/java/com/github/multimatum_team/multimatum/viewmodel/DeadlineListViewModel.kt
```diff
@@ -58,32 +59,38 @@ class DeadlineListViewModel @Inject constructor(
...
      */
-    fun addDeadline(deadline: Deadline, callback: (DeadlineID)->Unit) = viewModelScope.launch {
-        val id = deadlineRepository.put(deadline)
-        callback(id)
-    }
+    fun addDeadline(deadline: Deadline, callback: (DeadlineID) -> Unit = {}) =
+        viewModelScope.launch {
+            val id = deadlineRepository.put(deadline)
+            callback(id)
+        }
```
ValentinAebi, line 84: If I understand correctly, the goal of the callback argument is to be able to execute something in the `launch` block, but why is that needed? In what way is that different from simply returning the id and executing the callback after `addDeadline` returned?

felko: We cannot return from a `Job` (`viewModelScope.launch` returns a `Job` instance). `launch` is some kind of `fork` operation which returns immediately, it starts a task on the viewmodel thread and returns a reference to the job so that we can chose to cancel it for example, but it doesn't wait for the job to complete before returning. Also, unlike coroutines and `CompletableFuture`s, the `Job` class does not provide an equivalent to `addOnCompleteListener` so we have to handle callbacks ourselves.

app/src/main/java/com/github/multimatum_team/multimatum/viewmodel/DeadlineListViewModel.kt
```diff
@@ -58,32 +59,38 @@ class DeadlineListViewModel @Inject constructor(
...
 
     /**
      * Get all deadlines.
      */
     fun getDeadlines(): LiveData<Map<DeadlineID, Deadline>> =
         _deadlines
 
+    fun getDeadline(id: DeadlineID): Deadline =
+        _deadlines.value!![id]!!
```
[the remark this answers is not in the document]
felko: The first `!!` can fail if `_deadlines` is not yet initialized. The only way this can happen is that `getDeadline` is called immediately after the app started which is possible but unlikely. The second `!!` should never fail if the `id` exists, which should always be the case if we use the IDs provided by the repository.

felko:
> every things seems good. And just to be sure. The callback you use in DeadlineListViewModel are here to works asynchronously with furebase that's right ?

@Voxanimus Yes, the callbacks are executed as soon as the repository operations complete.

ValentinAebi (approved):

Voxanimus (approved):

PR #170, opened 2022-04-23
Feature/sort deadline

Voxanimus:


Matathiasq13824 (changes requested):
Seems fine to me, however, are you sure that the tests of the adapter do exactly what their name implies ?

app/src/main/java/com/github/multimatum_team/multimatum/model/DeadlineAdapter.kt
```diff
@@ -42,24 +42,33 @@ class DeadlineAdapter(private val context: Context) : BaseAdapter() {
...
-        dataSource = deadlines.toSortedMap { id1, id2 ->
-            compareValues(deadlines[id1]!!.dateTime, deadlines[id2]!!.dateTime)
-        }
+        dataSource = sortDeadline(deadlines)
+                // deadlines.toSortedMap { id1, id2 -> compareValues(deadlines[id1]!!.dateTime, deadlines[id2]!!.dateTime) }
         notifyDataSetChanged()
     }
 
+    private fun sortDeadline(deadlines: Map<DeadlineID, Deadline>): List<Pair<DeadlineID, Deadline>> {
+        val doneDeadlines: List<Pair<DeadlineID, Deadline>> = deadlines
```
felko, line 57: Instead of duplicating the code for each deadline state, consider doing something like:
```kt
    val partition: MutableMap<DeadlineState, MutableList<Pair<DeadlineID, Deadline>>> = DeadlineState
        .values()
        .associate { state -> Pair<DeadlineState, MutableList<Pair<DeadlineID, Deadline>>>(state, mutableListOf()) }
       	.toMutableMap()
      
    for ((id, deadline) in deadlines.entries) {
        partition[deadline.state]!!.add(Pair(id, deadline))
    }
    
    val result: MutableList<Pair<DeadlineID, Deadline>> = mutableListOf()
    for (state in DeadlineState.values()) {
        result.addAll(partition[state]!!.sortedBy { it.second.date })
    }
```
This way the code will continue working even if we add new deadline states.

felko (changes requested):

app/src/main/java/com/github/multimatum_team/multimatum/model/DeadlineAdapter.kt
Voxanimus: can you explain me briefly what this code do. I'm not sur to understand why we need to use mutableMap

felko: This snippet groups deadlines by states, then concatenates the groups sorted by dates. Doing so avoids traversing the list of deadlines twice, and will work even if we define new deadline states other than just todo and done.
You're right that we don't need the MutableMap for partition, an immutale Map is enough.

felko (approved):

Matathiasq13824 (approved):

PR #174, opened 2022-04-26
Logging/disable logging when not debug

ValentinAebi:
The methods in `LogUtil` involve reflection and must perform a few operations before being able to log something. Therefore, it is better for efficiency to disable them when building the release version of the app. This is implemented using two versions of `LogUtilImpl`, one in the `debug` directory that does actual logging, and another version in the `release` directory that does nothing. Methods in `LogUtil` simply call the equivalent method in the version of `LogUtilImpl` that has been included in the build.

### DIFF
--- a/app/src/release/java/com/github/multimatum_team/multimatum/LogUtilImpl.kt
+++ b/app/src/release/java/com/github/multimatum_team/multimatum/LogUtilImpl.kt
@@ -1,0 +1,16 @@
+package com.github.multimatum_team.multimatum
+
+/**
+ * Release implementation of LogUtil functions (do nothing in release mode)
+ */
+class LogUtilImpl: LogUtil.FunctionsProvider {
+
+    override fun debugLog(str: String) = doNothing()
+
+    override fun logFunctionCall() = doNothing()
+
+    override fun logFunctionCall(str: String) = doNothing()
+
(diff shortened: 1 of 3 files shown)


app/src/release/java/com/github/multimatum_team/multimatum/LogUtilImpl.kt
```diff
@@ -0,0 +1,16 @@
+package com.github.multimatum_team.multimatum
+
+/**
+ * Release implementation of LogUtil functions (do nothing in release mode)
+ */
```
macintosh1398, line 5: Same here, it could be good to explain why we have two different classes.

macintosh1398 (changes requested):
Good job, these classes will be useful and the fact that the logs are now optimized for the release version is a very good thing. I have nothing to add for the code. However, try to be a bit more explicit in the description of the classes for a more "external" point of view.

macintosh1398 (approved):

Matathiasq13824 (approved):
Great PR. A lot of pertinent comments from @macintosh1398 so I don't have to say anything more. Just a little nitpick, It coulb be nice to begin every comments in the code with an uppercase but this is good otherwise

PR #175, opened 2022-04-26
Create ProcrastinationDetectorSensorListener

### LINKED ISSUE #157: Split ProcrastinationDetectorService into 2 files and adapt tests

ValentinAebi:
As reported by CodeClimate, the `ProcrastinationDetectorService` class was becoming really big, so I split it into two files, adding a new class, `ProcrastinationDetectorSensorListener` that handles the sensor events. Sensor events related methods should still be called on instances of `ProcrastinationDetectorService`, and these methods call the equivalent methods on an instance of `ProcrastinationDetectorSensorListener`, which is private to `ProcrastinationDetectorService`. This design may seem a bit weird, but it has the advantage of keeping the API of `ProcrastinationDetectorService` unchanged, which allows us to avoid modifying the tests. In particular, it avoids the need for a new injected attribute.

Closes #157

### DIFF
--- a/app/src/main/java/com/github/multimatum_team/multimatum/service/ProcrastinationDetectorSensorListener.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/service/ProcrastinationDetectorSensorListener.kt
@@ -1,0 +1,104 @@
+package com.github.multimatum_team.multimatum.service
+
+import android.content.Context
+import android.hardware.Sensor
+import android.hardware.SensorEvent
+import android.hardware.SensorEventListener
+import android.widget.Toast
+import com.github.multimatum_team.multimatum.R
+import kotlin.math.abs
+
+class ProcrastinationDetectorSensorListener(private val applicationContext: Context) :
+    SensorEventListener {
+
+    // the service will wait for this number of detections before starting to display toasts
+    private var nonReportedDetectionsCnt = NON_REPORTED_DETECTIONS_CNT_INIT
+
+    // data relative to the last time a movement was detected
+    private var lastDetectionTimestampNanos: Long = LAST_DETECTION_TIMESTAMP_NONINIT_CODE
+    private var lastPosition: Array<Float>? = null  // null only at initialization
+
+    override fun onAccuracyChanged(sensor: Sensor?, accuracy: Int) {
+        /* Nothing to do */
+    }
+
+    override fun onSensorChanged(event: SensorEvent?) {
+        requireNotNull(event)
+        val currentTimeNanos = event.timestamp
+        initLastDetectionIfNotYetInitialized(currentTimeNanos)
+        initLastPositionWithCurrentIfNull(event)
+        // check whether enough time has passed since the last detection (o.w. do nothing)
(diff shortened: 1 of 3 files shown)


macintosh1398 (approved):
I think it was a good thing to split the procrastinationDetector into two files. Honestly I have nothing to add. The functions are in the file they should belong to, we have no magic values (text, colors, ...), the code is well commented and everything is clear, good work.

app/src/main/java/com/github/multimatum_team/multimatum/service/ProcrastinationDetectorSensorListener.kt
```diff
@@ -0,0 +1,104 @@
...
+    // the service will wait for this number of detections before starting to display toasts
+    private var nonReportedDetectionsCnt = NON_REPORTED_DETECTIONS_CNT_INIT
+
+    // data relative to the last time a movement was detected
+    private var lastDetectionTimestampNanos: Long = LAST_DETECTION_TIMESTAMP_NONINIT_CODE
+    private var lastPosition: Array<Float>? = null  // null only at initialization
+
+    override fun onAccuracyChanged(sensor: Sensor?, accuracy: Int) {
+        /* Nothing to do */
+    }
```
Matathiasq13824, line 23: Could it be possible to make a test for this function ? (Just to call it to increase coverage based on codeclimate)

app/src/main/java/com/github/multimatum_team/multimatum/service/ProcrastinationDetectorService.kt
```diff
@@ -34,23 +32,25 @@ class ProcrastinationDetectorService : Service(), SensorEventListener {
...
-    private var lastDetectionTimestampNanos: Long = LAST_DETECTION_TIMESTAMP_NONINIT_CODE
-    private var lastPosition: Array<Float>? = null  // null only at initialization
+    private val binder = PdsBinder()
 
-    // the service will wait for this number of detections before starting to display toasts
-    private var nonReportedDetectionsCnt = NON_REPORTED_DETECTIONS_CNT_INIT
+    private lateinit var sensorListener: ProcrastinationDetectorSensorListener
 
-    private val binder = PdsBinder()
+    override fun onAccuracyChanged(sensor: Sensor?, accuracy: Int) =
```
Matathiasq13824, line 39: Could it be possible to make a test for this function too ? (Just to call it to increase coverage based on codeclimate)

Matathiasq13824 (changes requested):
Great job. Except the two little comments that I have, You should add comments on the private functions of `ProcrastinationDetectorSensorListener` so that any people can easily see what everything do if they have to change something.

app/src/main/java/com/github/multimatum_team/multimatum/service/ProcrastinationDetectorSensorListener.kt
ValentinAebi: Does that even count in coverage? There is no code in it!

Matathiasq13824: Infortunatly, based on codeClimate, yes

Matathiasq13824 (approved):

Issue #169, opened 2022-04-23
Enable group administrators to invite people to their group


felko:
Group owners should be able to invite people to join their group. I think the simplest solution is to use [Firebase email action links](https://firebase.google.com/docs/auth/admin/email-action-links) which does not require any new activity.

Issue #168, opened 2022-04-23
Implement activity to create user groups


felko:
To enable users to share deadlines, users should be able to create groups.

PR #179, opened 2022-04-26
First draft of the Firebase backend for the user group feature

### LINKED ISSUE #176: Define a model for user groups
To start working on user groups, we should first agree on what defines a group. I think a group should be defined by:
- A unique identifier
- A name
- A single group administrator (though we could have multiple admis eventually)
- A list of members in that group (including the owner)

### LINKED ISSUE #177: Implement a repository for groups
Once we have defined a model for user groups (#176), we should have a way to store them on Firebase and fetch informations about the groups we administrate.

### LINKED ISSUE #189: Implement firestore mocking to test `FirebaseGroupRepository`
As suggested by @ValentinAebi, we can in fact test Firebase-dependent code using mockito, which will greatly improve our coverage at the cost of basically reimplementing all the Firebase functions that we are using. This issue focuses on `FirebaseGroupRepository` but we can ultimately use the same technique for `FirebaseDeadlineRepository` and `FirebaseAuthRepository`.

felko:
This PR defines user groups (closes #176) as well as a repository to store data about user groups in the Firebase database (closes #177).
The definition of user group and the interface of the group repository are to be refined in the future, when we will implement the frontend for this feature (#168, #169).
As usual with Firebase-dependent code, I was not able to test much, so expect a significant drop in coverage.

Since we managed to implement testing, this PR closes #189 as well.


app/src/main/java/com/github/multimatum_team/multimatum/model/User.kt
```diff
@@ -2,10 +2,23 @@ package com.github.multimatum_team.multimatum.model
 
 typealias UserID = String
 
+/**
+ * Model for users.
+ * @param id an unique identifier provided by the user repository
+ */
 sealed class User(open val id: UserID)
```
ValentinAebi, line 9: Should one really be able to instantiate `User`? Would it not be better to make it an interface?

app/src/main/java/com/github/multimatum_team/multimatum/model/UserGroup.kt
```diff
@@ -0,0 +1,29 @@
...
+data class UserGroup(
+    val id: GroupID,
+    val name: String,
+    val owner: UserID,
+    val members: Set<UserID>
+) {
+    init {
+        if (!members.contains(owner)) {
+            throw IllegalArgumentException()
+        }
```
ValentinAebi, line 24:
```suggestion
        require(members.contains(owner))
```

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseGroupRepository.kt
```diff
@@ -0,0 +1,108 @@
...
+import com.google.firebase.firestore.FirebaseFirestore
+import com.google.firebase.firestore.ktx.firestore
+import com.google.firebase.ktx.Firebase
+import kotlinx.coroutines.tasks.await
+
+/**
+ * Remote Firebase repository for storing user groups.
+ */
+class FirebaseGroupRepository : GroupRepository() {
+    private var database: FirebaseFirestore = Firebase.firestore
```
ValentinAebi, line 16: Couldn't `database` be a `val`?

app/src/test/java/com/github/multimatum_team/multimatum/util/MockGroupRepository.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    private val groups: MutableMap<GroupID, UserGroup>
+        get() = groupsPerUser.getOrPut(_user.id) { mutableMapOf() }
+
+    private val updateListeners: MutableMap<UserID, MutableList<(Map<GroupID, UserGroup>) -> Unit>> =
+        mutableMapOf()
+
+    init {
+        for (group in initialContents) {
+            val newID = (counter++).toString()
```
ValentinAebi, line 31: Using the return value of a `++`operator (either prefix or postfix) is confusing.
```suggestion
            val newID = counter.toString()
            counter++
```

app/src/test/java/com/github/multimatum_team/multimatum/util/MockGroupRepository.kt
```diff
@@ -0,0 +1,62 @@
...
+        setUser(AnonymousUser("0"))
+    }
+
+    private fun notifyUpdateListeners() =
+        updateListeners[_user.id]?.forEach { it(groups) }
+
+    override suspend fun fetchAll(): Map<GroupID, UserGroup> = groups
+
+    override suspend fun create(name: String): GroupID {
+        val id = (counter++).toString()
```
ValentinAebi, line 47:
```suggestion
        val id = counter.toString()
        counter++
```

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseGroupRepository.kt
ValentinAebi: Also, you say that it is almost impossible to test some parts of the code. Wouldn't it be possible to inject `database` in this class, so that we could generate a mock database using Mockito? If it works it would fix a large part of the problem as this class has 0% coverage.

ValentinAebi (changes requested):

app/src/main/java/com/github/multimatum_team/multimatum/model/User.kt
felko: Sealed classes cannot be instantiated, also interfaces cannot define attributes

app/src/main/java/com/github/multimatum_team/multimatum/model/UserGroup.kt
felko: good call, thanks

app/src/main/java/com/github/multimatum_team/multimatum/model/User.kt
ValentinAebi: Indeed it cannot be instantiated, my bad. But I still find it a bit weird to use a class for this purpose. And interfaces can define attributes, at least
```Kotlin
sealed interface Foo {
    val value: Int
}
data class Bar(override val value: Int): Foo
```
compiles.

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseGroupRepository.kt
felko: Possibly, it could increase the coverage but wouldn't actually test anything meaningful. Another more concerning issue is that, from what I understand, mockito only lets you specify what is returned from certain method calls of the mocked object, e.g. `when(mockedObject.method(arg1, arg2)).thenReturn(result)` but in our case we have to mock chains of method calls: consider the `fetch` implementation in `FirebaseGroupRepository` (which is the simplest method in the class):
```kt
    override suspend fun fetch(id: GroupID): UserGroup =
        deserializeGroup(database.collection("groups").document(id).get().await())
```
If we were to mock `FirebaseFirestore`, then wouldn't we have to do something like this?
```ks
val mockSnapshot = mock(DocumentSnapshot::class.java)
when(mockSnapshot.get("name")).thenReturn("test name")
... // same for owner and members
val mockDocument = mock(DocumentReference::class.java)
when(mockDocument.get()).thenReturn(mockSnapshot)
val mockCollection = mock(CollectionReference::class.java)
when(mockCollection.document(dummyID)).thenReturn(mockDocument)
val mockDatabase = mock(FirebaseFirestore::class.java)
when(mockDatabase.collection("groups")).thenReturn(mockCollection)
```

ValentinAebi: Yes I think that you would have to do that. I am not sure of how many such "chains" of mock you would have to do, maybe it is feasible (assuming one person is tasked with this and if needed spends a whole sprint on it)? Because achieving 80% coverage is not easy and we cannot afford that a whole large class is not tested at all... Otherwise you may be able to do some (ugly, but meaningful) tests using reflection?

felko: I wrote a draft and it indeed seems to be working, but is extremely tedious. Perhaps we can automate the process by mocking the entire firebase database from a hashmap?

ValentinAebi: This would probably make a lot of sens, yes

felko: I got a 11% increase with my dumb test, I think I can reach 50% without too much trouble.

app/src/main/java/com/github/multimatum_team/multimatum/model/User.kt
felko: Ah, I didn't know Kotlin could do that. What would be the benefit of defining an interface rather than a class in this case?

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseGroupRepository.kt
felko: I wrote some helper functions to generate the mock firebase database as discussed above, and managed to reach 44% diff-coverage by testing `fetchAll`. Unfortunately, I'm not sure I see how to go further and mock `add` and `delete`, which require changing the state of the database. I don't see a way to mock `addSnapshotListener` either.
I'll try to investigate how to implement this tomorrow, but if I don't succeed, maybe we can still remove the 50% threshold temporarily, and I'll make up for the -1.5% total coverage by applying the same technique for FirebaseDeadlineRepository.

app/src/main/java/com/github/multimatum_team/multimatum/model/User.kt
ValentinAebi: It would make code clearer (at least in my opinion). In the subclasses it is weird to override `id` and simultaneously pass it to the constructor of the supertype.

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseGroupRepository.kt
ValentinAebi: I don't exactly get the problem. What `add` and `delete` do you need to mock?

felko:
Since we managed to implement testing, this PR will also close #189.

ValentinAebi (approved):

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseGroupRepository.kt
felko: > I don't exactly get the problem. What `add` and `delete` do you need to mock?

The ones from Firebase, specifically `CollectionReference.add` and `DocumentReference.delete`

PR #180, opened 2022-04-27
Refactoring darkmode and finishing to apply it

macintosh1398:
Refactoring xml files related to darkmode. Every color used in the app are now centralized in the colors files. The "deadlineDetailsActivity" has been updated to support the darkmode (some colors are selected according to the shared preferences) and can be easily updated if new changes are made. Some icons attributes were also refactored.


LennyDZ (changes requested):
Just be coherent in the xml wether you use `@android:color`or `@color`to refer to the file.

app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
```diff
@@ -228,6 +238,18 @@ class DeadlineDetailsActivity : AppCompatActivity() {
 
     }
 
+    // Change the color of the date and title views according to the theme
+    private fun adaptToCurrentTheme() {
+        val isNightMode = sharedPreferences.getBoolean(MainSettingsActivity.DARK_MODE_PREF_KEY, false)
```
ValentinAebi, line 243: This means that if no value is found in `sharedPreferences` then dark mode is disabled right? I think that ideally the `false` default value should be a constant, named e.g. `IS_DEFAULT_DARK_MODE_ENABLED`.

app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
```diff
@@ -228,6 +238,18 @@ class DeadlineDetailsActivity : AppCompatActivity() {
...
+        var textColor = Color.WHITE
+
+        if(editMode) textColor = Color.BLACK
```
ValentinAebi, line 246:
```suggestion
        val textColor = if (editMode) Color.BLACK else Color.WHITE
```

ValentinAebi (changes requested):

app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
macintosh1398: Good point

ValentinAebi (approved):

LennyDZ (approved):

PR #187, opened 2022-04-27
Feature/date time recognition

### LINKED ISSUE #160: Create a system to recognize time information like "at 10am" in a deadline title

ValentinAebi:
This PR contains the implementation of a parser that is intended to find date and time information in deadline titles. Currently it is able to parse time information, but not dates (this should be implemented in a future sprint). The parser first splits the string into tokens (more or less equivalent to words), and then tries to find a matching pattern among a list of patterns (implemented as functions). If a pattern is found, a time (or, in the future, a date) is created accordingly.

Currently only the logic is implemented, it is not actually linked to the app yet.

Closes #160

### DIFF
--- a/app/src/main/java/com/github/multimatum_team/multimatum/model/datetime_parser/DateTimeExtractor.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/model/datetime_parser/DateTimeExtractor.kt
@@ -6,46 +6,138 @@ import java.util.*
 object DateTimeExtractor {
 
     /*
+        TODO implement date recognition + more patterns for time recognition
+
         When finished, should be able to parse:
 
         18h00
         18h
         6am
         6pm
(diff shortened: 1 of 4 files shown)


felko (approved):
Looks good, but it's hard to judge until the feature is finalized.
Maybe it's a bit late but I think using regexes would have been much simpler.

ValentinAebi:
> Looks good, but it's hard to judge until the feature is finalized. Maybe it's a bit late but I think using regexes would have been much simpler.

I also thought of using regexes, but I am not sure that they would really make it simpler. I decided to implement it from scratch because I want the parsing system to have maximum flexibility, and I think that to achieve this level of flexibility with regexes we would end up with code that is as complex as the current implementation. And from my experience regexes often lead to bugs that are hard to deal with.

PR #188, opened 2022-04-28
Modified tests for robolectric

Matathiasq13824:
Last week, we had a problem with the stability of some tests in CI, leading to the removal of two tests.
Using Robolectric, I have tried to fix them, which has moved `MainActivityTest` in the unit test folder.
I think that there is still room for improvement for the coverage but I was the best I could do with the time allocated.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -57,12 +57,12 @@ dependencies {
     implementation 'com.google.android.material:material:1.5.0'
     implementation 'androidx.constraintlayout:constraintlayout:2.1.3'
     implementation platform('com.google.firebase:firebase-bom:29.2.1')
-    implementation 'com.google.firebase:firebase-firestore-ktx'
-    implementation 'com.google.firebase:firebase-auth:21.0.3'
-    implementation 'com.firebaseui:firebase-ui-auth:7.2.0'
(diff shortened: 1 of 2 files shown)


ValentinAebi (commented):
`grantPremission` and `denyPermission look like they have the same code. Why is it? That is very weird since they have different roles.

felko (approved):
Thank you for fixing this, I have nothing to add other than what @ValentinAebi said.

Matathiasq13824:
> `grantPremission` and `denyPermission` look like they have the same code. Why is it? That is very weird since they have different roles.

The code before was also very similar. The most important part are the 'shadowApplication.grantPermissions' and `shadowApplication.denyPermissions` who will give or not the permission to access the camera

Voxanimus (approved):
It will be so usefull thanks

ValentinAebi:
> The code before was also very similar. The most important part are the `shadowApplication.grantPermissions` and `shadowApplication.denyPermissions` who will give or not the permission to access the camera

Yes but you never call `shadowApplication.denyPermissions`. You use `shadowApplication.grantPermissions` in both `grantPermission` and `denyPermission`

PR #209, opened 2022-05-03
Feature/pdf

LennyDZ:
This add feature to link pdf to deadline.

For now, only provide a way to browse and select pdf in addDeadline


app/src/main/java/com/github/multimatum_team/multimatum/activity/AddDeadlineActivity.kt
```diff
@@ -159,4 +175,44 @@ class AddDeadlineActivity : AppCompatActivity() {
...
+    }
+
+    //handle result of startActivity
+    val startForResult =
+        registerForActivityResult(ActivityResultContracts.StartActivityForResult()) { result: ActivityResult ->
+            if (result.resultCode == Activity.RESULT_OK) {
+                val intent = result?.data!!
+                pdfUri = intent.data!!
+                val uriString: String = pdfUri.toString()
+                var pdfName: String? = null
```
ValentinAebi, line 192: Why this declaration here? `pdfName` could simply be a `val` declared when you initialize it I think.

app/src/main/java/com/github/multimatum_team/multimatum/activity/AddDeadlineActivity.kt
```diff
@@ -36,6 +43,8 @@ class AddDeadlineActivity : AppCompatActivity() {
     private val deadlineListViewModel: DeadlineListViewModel by viewModels()
     private lateinit var textDate: TextView
     private lateinit var textTime: TextView
+    private lateinit var pdfTextView: TextView
+    private var pdfUri: Uri = Uri.EMPTY
```
macintosh1398, line 47: For clarity, it would be good to reorder vals and vars in this file. It means put lateinit together, and val/var by type.

macintosh1398 (commented):
In general the code looks good. Maybe just check the variable/value declaration order for clarity. Adding a little bit more comment in the code could also be good for external readers.

app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
```diff
@@ -54,6 +55,8 @@ class DeadlineDetailsActivity : AppCompatActivity() {
     private var dateTime: LocalDateTime = LocalDateTime.of(2022, 10, 10, 10, 10)
     private var state: DeadlineState = DeadlineState.TODO
 
+    private lateinit var deadline: Deadline
```
Matathiasq13824, line 58: Why these changes ? It change nothing to the activity

app/src/main/res/layout/activity_add_deadline.xml
```diff
@@ -128,61 +111,121 @@
...
+                app:layout_constraintEnd_toEndOf="parent"
+                app:layout_constraintStart_toStartOf="parent"
+                app:layout_constraintTop_toBottomOf="@+id/textView2" />
 
-        <CheckBox
-            android:id="@+id/radio_notification_3d"
-            android:layout_width="wrap_content"
-            android:layout_height="wrap_content"
+            <TextView
+                android:id="@+id/textView2"
```
Matathiasq13824, line 202: Change the id of this textView to be able to understand what it is used for

Matathiasq13824 (changes requested):
I admit, I don't understand why there is so much changes in `activtity_add_deadline.xml`.
Maybe it is just github that is lost, but it seems that you add two constraints and two textView ? And why multiple bias, height and weight have changed ?

app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
LennyDZ: yeah but during unsuccessful try to add a pdf view, I had to have access to the deadline. And. I think it makes sense to have the deadline stored somewhere in a variable since the whole activity is about this deadline.

Matathiasq13824: You only need the pdf, you could have add a variable pdf to recuperate the information.  The deadline is recuperated only in `setDeadlineObserver()` as we recuperate the necessary informations in other variables. Each variables can be changed at a different time so It is not a good idea to keep all of it in a Deadline (because each time you change something, you would have to make a copy of the Deadline with the new variables wich is not efficient)
And, as we only need the full deadline here, we dont need to store it)
Also, as you tests weren't successful, it is always better to remove all unnecessary modifications

app/src/test/java/com/github/multimatum_team/multimatum/AddDeadlineTest.kt
```diff
@@ -1,14 +1,21 @@
 package com.github.multimatum_team.multimatum
 
+import android.app.Activity
```
Matathiasq13824, line 3: Remove unsued import

app/src/main/java/com/github/multimatum_team/multimatum/activity/AddDeadlineActivity.kt
```diff
@@ -57,40 +61,14 @@ class AddDeadlineActivity : AppCompatActivity() {
         findViewById<CheckBox>(R.id.radio_notification_3d).text =
             getString(R.string.notify_before, "3 days")
         selectedDate = clockService.now().truncatedTo(ChronoUnit.MINUTES)
-        updateDisplayedDateAndTime()
```
Matathiasq13824, line 60: This was a feature from another pr, you should keep it

Issue #178, opened 2022-04-26
Protect user group permissions in using Firebase rules


felko:
Similar to the rules that were added by #126 to protect deadlines from being accessed to unauthorized users, groups should benefit from the same protection.

PR #210, opened 2022-05-03
Group ViewModel

### LINKED ISSUE #190: Test `FirebaseDeadlineRepository` using mockito
After @ValentinAebi's suggestion to use mockito to test the `FirebaseGroupRepository` class, we managed to increase the coverage of PR #179. We could implement a similar thing to test `FirebaseDeadlineRepository`.

### LINKED ISSUE #191: Implement a viewmodel for user groups
Now that user groups have a corresponding repository (#179), we should next implement a viewmodel to interact with it from the UI (which will be needed for #182).

### LINKED ISSUE #212: Fix Firebase `SnapshotListener` duplication bug
When listening on database updates, the `FirebaseDeadlineRepository` adds `SnapshotListener`s with some callback. When the authentication state changes, a second listener is added on top of the previous one, causing the new signed-in user to receive updates from their previous account. This caused some latency and glitches in the deadline list view.

felko:
This PR implements the group viewmodel (closes #191).

Some refactors were required to implement it correctly:
- The deadline model now contains an owner attribute which can either be `UserOwned`, meaning that the deadline is personal, and `GroupOwned(groupID)` meaning that the deadline is shared between all members of the group identified by its ID.
- I also removed `notificationTimes` from the deadline model since it seems that it is not used anymore (@LennyDZ please confirm this is the case), since I changed the firebase format, we might as well do this change now
- The auth repository was refactored to fix the bug described below
- The deadline repository was partly rewritten to take groups into account, we can now see deadlines from groups as well as personal deadlines
- The viewmodels are updated and (hopefully) correctly linked together, each updating dependent repositories
- The JSON serializer was refactored to avoid serializing the deadline owner into the QR code which we don't want (@Voxanimus tell me if you agree with the changes)

I also found a few bugs for which I was too lazy to make a separate PR:
- In the deadline repository, listeners were not detached when the user logged out, so multiple listeners were observing firebase changes and updating the viewmodels concurrently which made the list invalid or glitchy (closes #212)
- In the deadline details activity, the modification was overwriting deadline attributes (we didn't notice it before since it was only overwriting notificationTimes and description which weren't used at the moment)
- The auth repository was creating anonymous users even when not necessary

Things to do before merging the PR:
- [x] The deadline repository implementation is currently compatible with the previous dealdine format stored in firebase, but we should remove the extra code at the end of this sprint.
- [x] Add a bunch of methods in the group repository and group viewmodel so that we can work at the same time on the group UI (#168)
- [x] Add tests (closes #190)

Due to the complications introduced by groups, the protection through firebase rules is currently disabled, something we should remedy in #178.

Sorry for the huge diff, I had no other choice than to do #190 in this PR to reach the 50% diff-coverage threshold.


app/src/main/java/com/github/multimatum_team/multimatum/ProfileActivity.kt
```diff
@@ -38,19 +40,23 @@ class ProfileActivity : AppCompatActivity() {
...
-    private fun updateUI(user: User) {
+    private fun updateUI(user: SignedInUser) {
         loginMessage.text =
-            getString(R.string.you_re_logged_in_as, (user as SignedInUser).email)
+            getString(R.string.you_re_logged_in_as, user.email)
     }
 
     companion object {
-        private const val TAG = "ProfileFragment"
+        private const val TAG = "ProfileActivity"
```
ValentinAebi, line 60: Is that for logging? That is just a suggestion, but you could maybe use the functions from `LogUtil` to make logging more practical

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseAuthRepository.kt
```diff
@@ -30,39 +54,15 @@ class FirebaseAuthRepository @Inject constructor() : AuthRepository {
...
-            Log.d(TAG, "First time log-in, waiting for anonymous sign-in...")
-            user = anonymousSignIn()
-            Log.d(TAG, "Successfully signed-in anonymously with user ID ${user.id}")
-        } else {
-            // Otherwise, we check whether the user has signed in using a real account or if they
-            // were signed in as an anonymous user
-            user = convertFirebaseUser(firebaseUser)
+    private suspend fun anonymousSignIn(): AnonymousUser =
+        when (val authResultUser = auth.signInAnonymously().await()?.user) {
+            null ->  throw RuntimeException("Failed to sign-in anonymously")
```
ValentinAebi, line 59: It would probably be better to replace `RuntimeException` by a more precise type of exception.

app/src/androidTest/java/com/github/multimatum_team/multimatum/util/MockAuthRepository.kt
```diff
@@ -11,36 +11,35 @@ import com.github.multimatum_team.multimatum.repository.DeadlineRepository
...
+    init {
+        _user = generateNewAnonymousUser()
+    }
 
     private val updateListeners: MutableList<(User) -> Unit> =
         mutableListOf()
 
     private fun notifyUpdateListeners() =
-        updateListeners.forEach { it(user) }
+        updateListeners.forEach { it(_user) }
```
Matathiasq13824, line 25: Why the change of name ?

app/src/main/java/com/github/multimatum_team/multimatum/activity/MainActivity.kt
```diff
@@ -18,8 +18,12 @@ import androidx.core.content.ContextCompat
 import com.github.multimatum_team.multimatum.R
 import com.github.multimatum_team.multimatum.adaptater.DeadlineAdapter
 import com.github.multimatum_team.multimatum.repository.DeadlineRepository
+import com.github.multimatum_team.multimatum.repository.GroupRepository
 import com.github.multimatum_team.multimatum.util.DeadlineNotification
 import com.github.multimatum_team.multimatum.viewmodel.DeadlineListViewModel
+import com.google.firebase.auth.FirebaseAuth
+import com.google.firebase.auth.FirebaseAuthSettings
+import com.google.firebase.firestore.FirebaseFirestore
```
Matathiasq13824, line 26: I only see in this activity that you have removed two empty line. Are these imports still useful ?

app/src/main/java/com/github/multimatum_team/multimatum/viewmodel/GroupViewModel.kt
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * ViewModel for the group view.
+ * Acts as a bridge between the group repository (Firebase) and the user interface.
+ * The ViewModel ensures that the data is always fresh, keeping the database and the UI in sync.
+ * The list of groups is also updated when the GroupRepository notifies that the authenticated
+ * user has changed.
+ */
+@HiltViewModel
+class GroupViewModel @Inject constructor(
```
Matathiasq13824, line 26: This class is completly untested. I don't think it is a good idea to make the tests this sprint with the work you have already done but it can be a good issue for a next sprint

Matathiasq13824 (changes requested):
As always, amazing work ! It was not easy and you have made, in addtion, some cleanup in every file and some tests. I see that some function are not yet tested but the coverage is good enough and this can wait another sprint.
I will approve later when all the tasks for this PR are done otherwise not really more to add

app/src/androidTest/java/com/github/multimatum_team/multimatum/util/MockAuthRepository.kt
felko: This is because the AuthRepository was refactored. Instead of creating an anonymous user in the `getCurrentUser()` method, there is now only a `getUser()` getter which returns the underlying `_user` that is set in the constructor.

app/src/main/java/com/github/multimatum_team/multimatum/activity/MainActivity.kt
felko: Good catch

app/src/main/java/com/github/multimatum_team/multimatum/viewmodel/GroupViewModel.kt
felko: I will test it in this PR, it's not too complicated (mostly the same as the other viewmodel tests)

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseAuthRepository.kt
felko: What do you suggest? This snippet was copied from the Firebase docs.

Voxanimus (changes requested):

LennyDZ:
I confirm that notificationTimes are not used anymore in firebase

Matathiasq13824 (approved):

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseAuthRepository.kt
ValentinAebi: `IllegalStateException`? The sign-in fails, which should not happen and now the repository is in a state that it cannot handle

app/src/androidTest/java/com/github/multimatum_team/multimatum/util/MockGroupRepository.kt
```diff
@@ -56,6 +56,14 @@ class MockGroupRepository(initialContents: List<UserGroup>) : GroupRepository()
         groups.remove(id)
     }
 
+    override suspend fun rename(id: GroupID, newName: String) {
+        groups[id] = groups[id]!!.copy(name = newName)
```
ValentinAebi, line 60: `rename` is public, so it should check its arguments and in particular throw an `IllegalArgumentException` if `groups[id] == null`, not a `NullPointerException`. So I think that a `require` is needed here.

app/src/androidTest/java/com/github/multimatum_team/multimatum/util/MockGroupRepository.kt
```diff
@@ -56,6 +56,14 @@ class MockGroupRepository(initialContents: List<UserGroup>) : GroupRepository()
...
+    }
+
+    override suspend fun invite(id: GroupID, email: String) {
+        throw NotImplementedError("MockGroupViewModel.invite is not implemented")
```
ValentinAebi, line 64: Sorry to insist on exceptions, but `NotImplementedError` suggests that the method is not *yet* implemented. Here `UnsupportedOperationException` would be clearer in my opinion.

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseGroupRepository.kt
```diff
@@ -94,6 +94,27 @@ class FirebaseGroupRepository @Inject constructor(database: FirebaseFirestore) :
...
+            .await()
+    }
+
+    /**
+     * Invite an user to join a group.
+     * @param id the ID of the grop to which we want to invite the user
+     * @param email the email of the user to invite
+     */
+    override suspend fun invite(id: GroupID, email: String) {
+        throw NotImplementedError("FirebaseGroupRepository.invite is not implemented")
```
ValentinAebi, line 115: Same comment as above about the exception type

app/src/main/java/com/github/multimatum_team/multimatum/repository/GroupRepository.kt
```diff
@@ -48,9 +48,23 @@ abstract class GroupRepository {
...
+     * @param newName the new name of the group
+     */
+    abstract suspend fun rename(id: GroupID, newName: String)
+
+    /**
+     * Invite an user to a group.
+     * @param id the ID of the group to which we want to invite the new user
+     * @param email the email of the user to invite
+     */
+    abstract suspend fun invite(id: GroupID, email: String)
```
ValentinAebi, line 67: Or maybe you could simply implement the method (i.e. throw the exception) here instead of delegating to subclasses?

app/src/test/java/com/github/multimatum_team/multimatum/util/MockGroupRepository.kt
```diff
@@ -56,6 +56,14 @@ class MockGroupRepository(initialContents: List<UserGroup>) : GroupRepository()
         groups.remove(id)
     }
 
+    override suspend fun rename(id: GroupID, newName: String) {
+        groups[id] = groups[id]!!.copy(name = newName)
+    }
+
+    override suspend fun invite(id: GroupID, email: String) {
+        throw NotImplementedError("MockGroupViewModel.invite is not implemented")
```
ValentinAebi, line 64: Here too, the exception time is confusing in my opinion

app/src/main/java/com/github/multimatum_team/multimatum/viewmodel/UserViewModel.kt
```diff
@@ -37,7 +37,7 @@ class UserViewModel @Inject constructor(
         _user
 
     fun signOut() = viewModelScope.launch {
-        Log.d("UserViewModel", "signing out from ${_user.value}")
+        LogUtil.debugLog("signing out from ${_user.value}")
```
ValentinAebi, line 40: I see that you are using the methods from `LogUtil`, are they working correctly now?

ValentinAebi (changes requested):

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseAuthRepository.kt
felko: Looking at the documentation of `IllegalStateException`, it seems that it's rather thrown to signal that the method was called in an illegal state rather than the method leaving the program in an illegal state

app/src/main/java/com/github/multimatum_team/multimatum/viewmodel/UserViewModel.kt
felko: Yes, thank you for the fix

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseAuthRepository.kt
ValentinAebi: Indeed, ideally you could define a custom exception as an object but it's fine to keep it like that, the message does the job

app/src/androidTest/java/com/github/multimatum_team/multimatum/util/MockGroupRepository.kt
felko: Yeah that makes sense, I don't think any tests will be required on the invite feature.

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseGroupRepository.kt
felko: > NotImplementedError suggests that the method is not yet implemented.

This is exactly what I meant in this case

app/src/test/java/com/github/multimatum_team/multimatum/util/MockGroupRepository.kt
felko: I think the best option is to throw `UnsupportedOperationException` in the `MockGroupRepository` and `NotImplementedError` in the `FirebaseGroupRepository`.

app/src/main/java/com/github/multimatum_team/multimatum/repository/GroupRepository.kt
felko: I think the best option is to throw `UnsupportedOperationException` in the `MockGroupRepository` and `NotImplementedError` in the `FirebaseGroupRepository`.

app/src/androidTest/java/com/github/multimatum_team/multimatum/util/MockGroupRepository.kt
felko: This class is only used in tests but sure

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseGroupRepository.kt
ValentinAebi: Oh OK, I thought you would not implement it at all

app/src/main/java/com/github/multimatum_team/multimatum/repository/GroupRepository.kt
ValentinAebi: Makes sense in this case

ValentinAebi (approved):

Voxanimus (approved):

PR #219, opened 2022-05-04
Adapting darkmode to DeadlineListViewModel

macintosh1398:
This very small PR solves the display problem seen last week where the "due to" text is not displayed correctly in darkmode. The tests have been changed accordingly.

### DIFF
--- a/app/src/main/java/com/github/multimatum_team/multimatum/adaptater/DeadlineAdapter.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/adaptater/DeadlineAdapter.kt
@@ -138,7 +138,6 @@ class DeadlineAdapter(
             }
             clockService.now() > deadline.dateTime -> {
                 detail = context.getString(R.string.isAlreadyDue)
-                detailTextView.setTextColor(Color.BLACK)
                 detailTextView.setTypeface(detailTextView.typeface, Typeface.NORMAL)
             }
             else -> {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/github/multimatum_team/multimatum/adaptater/DeadlineAdapter.kt
```diff
@@ -138,7 +138,6 @@ class DeadlineAdapter(
             }
             clockService.now() > deadline.dateTime -> {
                 detail = context.getString(R.string.isAlreadyDue)
-                detailTextView.setTextColor(Color.BLACK)
```
Matathiasq13824, line 141: As we set the color of this text to green and red in another part of the code and the addition of the done button last week, we can have a problem.
Suppose that we have a deadline done who is already due. The color of this text should be green and show "Done". However, if the user press the done button, it will change the state of deadline to TODO and this text too.
As the deadline is already due, the text will noe be "Is already due" **but the text will still be in green** as we don't change the color. This setTextColor was to resolve this (but I didn't thought about the darkmode).
To make this work, I think you can put this instead
`detailTextView.setTextColor(getResources().getColor(R.color.deadline_details_title))`

macintosh1398: Yes I will try that. The function that you suggest is deprecated but I will try to do it with the actual equivalent.

Matathiasq13824 (approved):

felko (approved):

PR #220, opened 2022-05-04
Adding the location autocomplete feature for deadlines

macintosh1398:
This PR gives the user the ability to select a place for its deadlines using the Google Places Autocomplete Fragment. This fragment can be found in the addDeadlineActivity and (for now) returns elementary information about the chosen location when the user select one. The processing and the storage of these infos are the next step. IMPORTANT: Infortunately, the Autocomplete fragment requires billing informations to be able to use it. So for now it doesn't display anything.


macintosh1398:
NOTE: Due to the billing problem explained above, this PR includes only code that can be safely executed without any problems (in other word the code is operational), but the API request cannot succeed in the sense that Google does not return any response. But it still is a good way to see what we should get when this issue will be solved.

app/src/main/java/com/github/multimatum_team/multimatum/activity/AddDeadlineActivity.kt
```diff
@@ -93,6 +101,41 @@ class AddDeadlineActivity : AppCompatActivity() {
...
+        val autocompleteFragment =
+            supportFragmentManager.findFragmentById(R.id.autocomplete_fragment) as AutocompleteSupportFragment
+
+        autocompleteFragment
+            .setPlaceFields(listOf(Place.Field.ID, Place.Field.NAME)) //Returned fields
+            .setCountries(listOf("CH")) // Default startup country
+
+        autocompleteFragment.setOnPlaceSelectedListener(object : PlaceSelectionListener {
+            override fun onPlaceSelected(place: Place) {
+                // TO DO: Define here what we do with the returned place
```
ValentinAebi, line 129: `TO DO` would better be in a single word, so that IDEs detect it

app/src/main/java/com/github/multimatum_team/multimatum/activity/AddDeadlineActivity.kt
```diff
@@ -93,6 +101,41 @@ class AddDeadlineActivity : AppCompatActivity() {
...
+                Log.i(ContentValues.TAG, "Place: ${place.name}, ${place.id}")
+            }
+
+            override fun onError(status: Status) {
+                Log.i(ContentValues.TAG, "An error occurred: $status")
```
ValentinAebi, line 134: If that is really an error (in the sense of something that should normally not happen), it would be better to use another level of logging. IIRC `Log.i` means "information", probably another function (`Log.e`?) is more suited to reporting an error

ValentinAebi (changes requested):

app/src/main/java/com/github/multimatum_team/multimatum/activity/AddDeadlineActivity.kt
macintosh1398: Good point

LennyDZ (approved):
Seems good, however I'll need some test.

macintosh1398:
Yes it is planned.

Issue #234, opened 2022-05-10
Create activity to see group details


felko:
Once we implement #233, we should be able to display group members (using their names) in a new activity, similar to `DeadlineDetailsActivity` but for groups, accessible from the `GroupsActivity`.

PR #235, opened 2022-05-10
User repository

### LINKED ISSUE #233: Define repository to put and access public user information
To show group owner and members, we should have a way to access public user information (such as their names).
I suggest we create a new collection in our Firebase database, containing user information, and implement a repository to access it from our code.

felko:
This PR implements the user repository as described in #233.
Thanks to this new repository, public user information is now stored in firebase firestore, which lets us see user names from user IDs, which wouldn't have been possible otherwise.
The reason for this repository is to show the name of the group owner and members in the `GroupsActivity` and the upcoming `GroupDetailsActivity` (see #234).

Closes #233.

### DIFF
--- a/app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseUserRepository.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseUserRepository.kt
@@ -52,7 +52,6 @@ class FirebaseUserRepository @Inject constructor(database: FirebaseFirestore) : 
             .whereIn(FieldPath.documentId(), ids)
             .get()
             .await()
-            .toList()
             .map { deserializeUserInfo(it) }
 
     /**
--- a/app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseUserRepository.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseUserRepository.kt
@@ -1,0 +1,66 @@
+package com.github.multimatum_team.multimatum.repository
+
+import com.github.multimatum_team.multimatum.model.UserID
+import com.github.multimatum_team.multimatum.model.UserInfo
+import com.google.firebase.firestore.DocumentSnapshot
+import com.google.firebase.firestore.FieldPath
+import com.google.firebase.firestore.FirebaseFirestore
+import kotlinx.coroutines.tasks.await
+import javax.inject.Inject
+
+/**
+ * An interface to the user repository.
+ * This lets us get publicly available information about signed-in users.
+ */
+class FirebaseUserRepository @Inject constructor(database: FirebaseFirestore) : UserRepository() {
+    private val usersRef = database
+        .collection("users")
+
+    /**
+     * Convert user information into a hashmap, so that we can send the user data to Firebase.
+     */
+    private fun serializeUserInfo(userInfo: UserInfo): HashMap<String, *> =
+        hashMapOf(
+            "name" to userInfo.name,
+        )
+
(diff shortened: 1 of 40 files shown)


Matathiasq13824 (approved):
Great job !
Really, I have nothing to add. The code is well commented and correctly formatted.
I could discuss about the cleaning of thecode in other files than the ones requiring the user repository but it is something that would be done one day so everything is fine for me

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseUserRepository.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Get public information of multiple users at once.
+     */
+    override suspend fun fetch(ids: List<UserID>): List<UserInfo> =
+        usersRef
+            .whereIn(FieldPath.documentId(), ids)
+            .get()
+            .await()
+            .map { deserializeUserInfo(it) }
```
ValentinAebi, line 55: Extreme nitpicking:
```suggestion
            .map(::deserializeUserInfo)
```
(check the syntax if you want to apply it)

app/src/test/java/com/github/multimatum_team/multimatum/DateTimeExtractorTest.kt
```diff
@@ -37,7 +37,7 @@ class DateTimeExtractorTest {
     }
 
     @Test
-    fun `18h_is_parsed_correctly`(){
+    fun `18h_is_parsed_correctly`() {
```
ValentinAebi, line 40: Be careful when applying `ctrl+alt+L` to files you don't modifiy, you may create lots of conflicts for a very small benefit!

app/src/test/java/com/github/multimatum_team/multimatum/util/MockFirebaseAuth.kt
```diff
@@ -0,0 +1,8 @@
+package com.github.multimatum_team.multimatum.util
+
+import com.google.firebase.auth.FirebaseAuth
+import org.mockito.Mockito
+
+class MockFirebaseAuth {
+    val auth: FirebaseAuth = Mockito.mock(FirebaseAuth::class.java)
+}
```
ValentinAebi, line 8: Shouldn't this be an object?

app/src/test/java/com/github/multimatum_team/multimatum/DateTimeExtractorTest.kt
felko: Yes I mistakenly formatted the entire project, I agree its a bad idea but by the time I realized that, I had already modified a bunch of (relevant) files.

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseUserRepository.kt
felko: Nice, I tried `.map(this::deserializeUserInfo)` but it didn't seem to work.

app/src/test/java/com/github/multimatum_team/multimatum/util/MockFirebaseAuth.kt
felko: For now yes, there is no particular reason for `MockFirebaseAuth` to be a class, but it will definitely need to be a class once I figure out how to mock it properly (by that I mean reproduce Firebase behavior as I did for `MockFirebaseFirestore`)

ValentinAebi (approved):

PR #244, opened 2022-05-11
Adding location search feature

macintosh1398:
This new feature enable the user to select a location when creating a new deadline. The Mapbox API has been used, and provide the user an interface to enter a location, and the user will then have the choice between the found locations. When selected, the location name is displayed on the add deadline screen. In the future, the coordinates will be extracted and stored on Firebase.


felko (approved):
It's a shame that we weren't able to make the tests work, but from what I can see in the commented code, everything looks fine, good work.

macintosh1398:
Yes it would have been great if the tests worked. Thanks for the approval and I am terribly sorry about that.

ValentinAebi (approved):
We will have to retry to make the tests work but overall it looks good

PR #246, opened 2022-05-12
Feature/more patterns in parser

### LINKED ISSUE #218: Add missing patterns to the parser for dates and times

### LINKED ISSUE #243: Use annotations to mark patterns in parser

### LINKED ISSUE #245: Bugfix: pattern 15h00 not working

ValentinAebi:
Implemented:
* new patterns for the parser
   close #218
* patterns are now marked using an annotation instead of having to add them to a list (which is error-prone)
   close #243 
* bug with pattern "15h00" matching only "15h" is fixed (problem was that the parser was looking for "h" as a `SymbolToken`, but it is actually an `AlphabeticToken`)
   close #245

### DIFF
--- a/app/.gitignore
+++ b/app/.gitignore
@@ -1,2 +1,3 @@
 /build
 google-services.json
+*.ws.kts
--- a/app/src/main/java/com/github/multimatum_team/multimatum/model/datetime_parser/DateTimePatternsGenerator.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/model/datetime_parser/DateTimePatternsGenerator.kt
@@ -1,6 +1,5 @@
 package com.github.multimatum_team.multimatum.model.datetime_parser
 
-import com.github.multimatum_team.multimatum.LogUtil
 import java.time.*
 import java.time.temporal.TemporalAdjusters
 import kotlin.reflect.full.findAnnotation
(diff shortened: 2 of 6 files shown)


ValentinAebi:
As a "side effect" this PR also adds dependencies for worksheets, which can be helpful to debug or test some things without launching the app

app/src/main/java/com/github/multimatum_team/multimatum/model/datetime_parser/DateTimePatternsGenerator.kt
```diff
@@ -103,53 +161,81 @@ class DateTimePatternsGenerator(private val currentDateProvider: () -> LocalDate
...
+    private val `3h00pm` =
+        listOf(
+            Token::asHour12,
+            timeSeparator,
+            Token::asMinute,
+            pm
+        ) to { args: List<Any?> ->
+            timeFor(hour = args.getInt(0) + 12, minute = args.getInt(2))
+        }
+
```
Voxanimus, line 213: not sur to understand why you have private attrivute for speciffically 3h pm and 3h am

app/.gitignore
```diff
@@ -1,2 +1,3 @@
 /build
-google-services.json
\ No newline at end of file
+google-services.json
+*.ws.kts
```
LennyDZ, line 3: What is that?

ValentinAebi: Files ending in `.ws.kts` are Kotlin worksheets. So here it states that Git should ignore worksheets.

app/src/main/java/com/github/multimatum_team/multimatum/model/datetime_parser/DateTimePatternsGenerator.kt
ValentinAebi: These are different patterns, in the "am" case one can take the given hour directly, in the "pm" case it is needed to add 12 (e.g. 5pm = 17h).

LennyDZ (approved):
Seems fine for me

Voxanimus (approved):

PR #257, opened 2022-05-17
Add custom Notifications

### LINKED ISSUE #251: Add modularity to the notifications UI

Matathiasq13824:
This PR implements the possibility to add custom notification if the default ones aren't enough for the user.
The Dialog to select notifications has now a new button who will launch another dialog to select with a picker a new time.
I have also realised that the notifications who were modified in `DeadlineDetailsActivity` werent updated in the shared preference, it is now solved. 
Last but not the least, I had a problem of import that I solved

This close #251

### DIFF
--- a/app/src/main/java/com/github/multimatum_team/multimatum/activity/AddDeadlineActivity.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/activity/AddDeadlineActivity.kt
@@ -33,6 +33,8 @@ import com.github.multimatum_team.multimatum.viewmodel.GroupViewModel
 import com.google.firebase.ktx.Firebase
 import com.google.firebase.ktx.initialize
 import com.google.firebase.storage.FirebaseStorage
+import com.hamsa.twosteppickerdialog.OnStepPickListener
+import com.hamsa.twosteppickerdialog.TwoStepPickerDialog
 import dagger.hilt.android.AndroidEntryPoint
 import java.time.Duration
 import java.time.LocalDateTime
@@ -44,18 +46,6 @@ import javax.inject.Inject
  */
 @AndroidEntryPoint
 class AddDeadlineActivity : AppCompatActivity() {
-
-    companion object {
-        // Memorisation of which checkBox is selected for the notifications
-        private val notificationSelected = booleanArrayOf(false, false, false, false)
-        private val nameNotifications = arrayOf("1 hour", "5 hours", "1 day", "3 days")
-        private val checkBoxIdTime = mapOf(
-            "1 hour" to Duration.ofHours(1).toMillis(),
-            "5 hours" to Duration.ofHours(5).toMillis(),
-            "1 day" to Duration.ofDays(1).toMillis(),
-            "3 days" to Duration.ofDays(3).toMillis()
-        )
-    }
 
     @Inject
     lateinit var clockService: ClockService
@@ -77,6 +67,15 @@ class AddDeadlineActivity : AppCompatActivity() {
 
     private var pdfData = Uri.EMPTY
 
+    // Memorisation of which checkBox is selected for the notifications
+    private var notificationSelected = booleanArrayOf(false, false, false, false)
+    private var nameNotifications = arrayOf("1 hour", "5 hours", "1 day", "3 days")
+    private var timeNotifications = arrayOf(
+        Duration.ofHours(1).toMillis(),
+        Duration.ofHours(5).toMillis(),
+        Duration.ofDays(1).toMillis(),
+        Duration.ofDays(3).toMillis()
+    )
 
     // Memorisation of which group is selected for the deadline
     private var groupSelected = 0
@@ -288,6 +287,9 @@ class AddDeadlineActivity : AppCompatActivity() {
 
         // Set the name of the done button
         alertDialogBuilder.setPositiveButton(getString(R.string.done), null)
+        alertDialogBuilder.setNeutralButton("Add Custom Notification"){_, _ ->
+            addNotifications(view)
+        }
         alertDialogBuilder.create()
         alertDialogBuilder.show()
     }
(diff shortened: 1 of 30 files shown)


app/src/main/java/com/github/multimatum_team/multimatum/activity/AddDeadlineActivity.kt
```diff
@@ -414,8 +464,8 @@ class AddDeadlineActivity : AppCompatActivity() {
     // Recuperate the information on which notification must be set before returning it in an array
     private fun retrieveNotificationsTimes(): ArrayList<Long> {
         val res = ArrayList<Long>()
-        for (i in 0 until checkBoxIdTime.count()) {
-            if (notificationSelected[i]) res.add(checkBoxIdTime[nameNotifications[i]]!!)
+        for (i in 0 until timeNotifications.count()) {
+            if (notificationSelected[i]) res.add(timeNotifications[i])
         }
         return res
```
ValentinAebi, line 470:
```suggestion
        return timeNotifications.filterIndexed { idx, _ -> notificationsSelected[idx] }
```
(not 100% sure of the syntax but should be very close to that)

app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
```diff
@@ -58,10 +59,19 @@ class DeadlineDetailsActivity : AppCompatActivity() {
...
 
+    // Memorisation of which checkBox is selected for the notifications
+    private var notificationSelected: BooleanArray = booleanArrayOf(false, false, false, false)
+    private var nameNotifications = arrayOf("1 hour", "5 hours", "1 day", "3 days")
+    private var timeNotifications = arrayOf(
+        Duration.ofHours(1).toMillis(),
+        Duration.ofHours(5).toMillis(),
+        Duration.ofDays(1).toMillis(),
+        Duration.ofDays(3).toMillis()
+    )
```
ValentinAebi, line 73: If I am not mistaken this array is already created in `AddDeadlineActivity`. Do we really need to duplicate it here?

app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
```diff
@@ -243,6 +253,11 @@ class DeadlineDetailsActivity : AppCompatActivity() {
             )
             if (state == DeadlineState.DONE) {
                 DeadlineNotification.deleteNotification(id, this)
+            } else {
+                val listNotifications: List<Long> =
+                    timeNotifications.toList()
+                        .filter { i -> notificationSelected[timeNotifications.indexOf(i)] }
```
ValentinAebi, line 259:
```suggestion
                        .filterIndexed { idx, _ -> notificationSelected[idx] }
```

app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
```diff
@@ -329,6 +341,25 @@ class DeadlineDetailsActivity : AppCompatActivity() {
         }
     }
 
+    // Set the information of the notifications given by the deadline.
+    // If there are more notifications than the default ones, the function
+    // add them on the arrays to be able to show them in the Dialog.
+    private fun setNotifications(notifications: List<Long>) {
+        for (time in notifications) {
+            if (timeNotifications.contains(time)) {
+                notificationSelected[timeNotifications.indexOf(time)] = true
```
ValentinAebi, line 350: It is very probably not a problem here as no two elements in `notifications` have the same value, but be careful with the use of `indexOf`. Prefer a pattern like:
```Kotlin
for ((idx, time) in notifications.withIndex()){
     ...
             notificationsSelected[idx] = true
     ...
}
```

app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
```diff
@@ -367,8 +398,9 @@ class DeadlineDetailsActivity : AppCompatActivity() {
      */
     private fun textNotification(): String {
         val notifications =
-            nameCheckBox.filter { t -> notificationSelected[nameCheckBox.indexOf(t)] }
+            nameNotifications.filter { t -> notificationSelected[nameNotifications.indexOf(t)] }
```
ValentinAebi, line 401: Here too you should be able to use `filterIndexed` instead of this ugly `indexOf` :)

app/src/main/java/com/github/multimatum_team/multimatum/activity/AddDeadlineActivity.kt
Matathiasq13824: I didn't know that, thanks !

app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
Matathiasq13824: We need a duplicate if we want default options when we modify the deadline.  We could put it somewhere else and recuperate it but I think that would make the code less readable, but it is just my opinion.

ValentinAebi: I am not sure of how it works, but from what I see it is duplicate state and that is dangerous. Maybe you can store these values into an immutable list, and call `toArray` (or equivalent, I don't remember exactly) on this list each time you need an array. So you would have the default values in the list and no dependency between the arrays.

ValentinAebi (approved):

app/src/main/java/com/github/multimatum_team/multimatum/activity/AddDeadlineActivity.kt
```diff
@@ -412,11 +468,7 @@ class AddDeadlineActivity : AppCompatActivity() {
...
     // Recuperate the information on which notification must be set before returning it in an array
-    private fun retrieveNotificationsTimes(): ArrayList<Long> {
-        val res = ArrayList<Long>()
-        for (i in 0 until checkBoxIdTime.count()) {
-            if (notificationSelected[i]) res.add(checkBoxIdTime[nameNotifications[i]]!!)
-        }
-        return res
+    private fun retrieveNotificationsTimes(): List<Long> {
+        return timeNotifications.filterIndexed { idx, _ -> notificationSelected[idx] }
     }
```
Voxanimus, line 473: nice improvment of readability

Voxanimus (approved):

app/src/main/java/com/github/multimatum_team/multimatum/activity/AddDeadlineActivity.kt
Matathiasq13824: Yes, it's all thanks to Valentin who has showedme filterIndexed

PR #258, opened 2022-05-17
Group invites

felko:
This PR implements group invites using Firebase dynamic links. The way it works is that, when clicking on the invite button in the group details activity, a https link is generated and stored into your clipboard, so that you can send it to people you want to invite. When people click on this link on their phone, a new activity called `GroupInviteActivity` will show up and prompt the user who can then chose to accept the invitation (upon which they will be added to the group and redirected to the main activity), or deny it (upon which the application is closed).

To implement this feature, I had to change the semantics of `GroupRepository.fetch` which before was expected to fetch a group that the currently logged-in user has joined. Now, it can fetch any group (which further complicates firebase protection rules)
Therefore, the group viewmodel includes a new `fetchNewGroup` method which allows retrieving any group from the repository, and not just a group which we belong to (which is what `getGroup` does.

### DIFF
--- a/app/src/main/java/com/github/multimatum_team/multimatum/activity/GroupDetailsActivity.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/activity/GroupDetailsActivity.kt
@@ -1,16 +1,14 @@
 package com.github.multimatum_team.multimatum.activity
 
 import android.app.AlertDialog
-import android.content.Context
-import android.content.DialogInterface
-import android.content.Intent
-import android.content.SharedPreferences
+import android.content.*
 import android.os.Bundle
 import android.text.SpannableStringBuilder
 import android.view.MotionEvent
 import android.view.View
 import android.widget.Button
 import android.widget.TextView
+import android.widget.Toast
 import androidx.activity.viewModels
 import androidx.appcompat.app.AppCompatActivity
 import androidx.recyclerview.widget.LinearLayoutManager
@@ -76,6 +74,7 @@ class GroupDetailsActivity : AppCompatActivity() {
         // Initialize UI widgets
         initTextInput()
         initGroupMemberView()
+        initInviteButton()
         initLeaveButton()
     }
 
@@ -140,6 +139,24 @@ class GroupDetailsActivity : AppCompatActivity() {
     }
 
     /**
+     * Generate invite link for the current group and store it in a clipboard.
+     */
+    private fun initInviteButton() {
+        groupInviteButton.setOnClickListener {
+            groupViewModel.generateInviteLink(groupID) { inviteLink ->
+                val clipboard: ClipboardManager =
+                    getSystemService(CLIPBOARD_SERVICE) as ClipboardManager
+                val clip =
+                    ClipData.newPlainText("Invite link to ${groupNameView.text}", inviteLink.toString())
+                clipboard.setPrimaryClip(clip)
+                Toast
+                    .makeText(this, "Copied invite link to clipboard", Toast.LENGTH_SHORT)
+                    .show()
+            }
+        }
+    }
+
+    /**
      * Initialize groupDeleteOrLeaveButton with the right text and leave the group when clicked.
      */
     private fun initLeaveButton() {
@@ -170,11 +187,11 @@ class GroupDetailsActivity : AppCompatActivity() {
             AlertDialog.Builder(this)
                 .setMessage(getString(R.string.group_delete_confirmation_dialog))
                 .setPositiveButton(getString(R.string.group_delete_dialog_confirm)) { dialog, _ ->
-                        dialog.dismiss()
-                        groupViewModel.deleteGroup(groupID)
-                        finish()
+                    dialog.dismiss()
+                    groupViewModel.deleteGroup(groupID)
+                    finish()
                 }.setNegativeButton(getString(R.string.group_delete_dialog_cancel)) { dialog, _ ->
-                        dialog.dismiss()
+                    dialog.dismiss()
                 }.show()
         }
     }
--- a/app/src/main/java/com/github/multimatum_team/multimatum/activity/GroupInviteActivity.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/activity/GroupInviteActivity.kt
@@ -6,20 +6,24 @@ import android.widget.Button
 import android.widget.TextView
 import androidx.activity.viewModels
 import androidx.appcompat.app.AppCompatActivity
+import com.github.multimatum_team.multimatum.LogUtil
 import com.github.multimatum_team.multimatum.R
 import com.github.multimatum_team.multimatum.model.GroupID
 import com.github.multimatum_team.multimatum.model.SignedInUser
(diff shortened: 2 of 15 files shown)


app/src/main/java/com/github/multimatum_team/multimatum/activity/GroupDetailsActivity.kt
```diff
@@ -139,6 +138,24 @@ class GroupDetailsActivity : AppCompatActivity() {
...
+    private fun initInviteButton() {
+        groupInviteButton.setOnClickListener {
+            groupViewModel.generateInviteLink(groupID) { inviteLink ->
+                val clipboard: ClipboardManager =
+                    getSystemService(CLIPBOARD_SERVICE) as ClipboardManager
+                val clip =
+                    ClipData.newPlainText("Invite link to ${groupNameView.text}", inviteLink.toString())
+                clipboard.setPrimaryClip(clip)
+                Toast
+                    .makeText(this, "Copied invite link to clipboard", Toast.LENGTH_SHORT)
```
Matathiasq13824, line 153: You could extract the strings variables

app/src/test/java/com/github/multimatum_team/multimatum/util/MockGroupRepository.kt
```diff
@@ -62,23 +65,36 @@ class MockGroupRepository(initialContents: List<UserGroup>) : GroupRepository()
...
+            .authority("multimatum.page.link")
+            .appendQueryParameter("id", group.id)
+            .build()
+        return Uri.Builder()
+            .scheme("https")
+            .authority("multimatum.page.link")
+            .appendQueryParameter("sd", "Click this link to accept the invite")
+            .appendQueryParameter("st", "Join group ${group.name}")
+            .appendQueryParameter("apn", "com.github.multimatum_team.multimatum")
+            .appendQueryParameter("link", inviteLink.toString())
```
Matathiasq13824, line 81: You could extract these string variables

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseGroupRepository.kt
```diff
@@ -109,12 +116,42 @@ class FirebaseGroupRepository @Inject constructor(database: FirebaseFirestore) :
...
+        val group = fetch(id)
+        val inviteLink = Uri.Builder()
+            .scheme("https")
+            .authority("multimatum.page.link")
+            .appendQueryParameter("id", group.id)
+            .build()
+
+        val dynamicLink = Firebase.dynamicLinks.dynamicLink {
+            link = inviteLink
+            domainUriPrefix = "https://multimatum.page.link"
```
Matathiasq13824, line 132: You could extract these string variables

Matathiasq13824 (changes requested):
Great job ! The only nitpick that I can say is that you should extract some of the string variable but that's it.

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseGroupRepository.kt
felko: I left the link hardcoded for now but extracted the link title and description , I'm not sure exactly what kind of strings are expected to be string resources, I think string resources are mainly here to facilitate internationalization but I could be wrong.

app/src/test/java/com/github/multimatum_team/multimatum/util/MockGroupRepository.kt
felko: We would need to inject a context in the mock repository to be able to access string resources, I'm not sure how to do this.

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebaseGroupRepository.kt
felko: I resolved this problem (for the above comment too) by decoupling android dependencies from the repository, now the link title and description are passed as parameters.

Matathiasq13824 (approved):

app/src/main/java/com/github/multimatum_team/multimatum/activity/GroupInviteActivity.kt
```diff
@@ -0,0 +1,123 @@
...
+                                when (val groupID = link.getQueryParameter("id")) {
+                                    null -> initInvalidLink()
+                                    else -> initValidLink(groupID)
+                                }
+                            }
+                        }
+                    }
+                    .addOnFailureListener(this) { e ->
+                        initInvalidLink()
+                    }
```
ValentinAebi, line 75: Maybe you can extract function(s) in this code block? This would make the code more readable and the name of the function(s) could help understanding what the purpose of this code block is.

app/src/test/java/com/github/multimatum_team/multimatum/util/MockFirebaseDynamicLinks.kt
```diff
@@ -0,0 +1,33 @@
...
+import com.google.firebase.dynamiclinks.FirebaseDynamicLinks
+import com.google.firebase.dynamiclinks.PendingDynamicLinkData
+import com.google.firebase.dynamiclinks.internal.DynamicLinkData
+import org.mockito.Mockito
+import org.mockito.Mockito.`when`
+import org.mockito.kotlin.any
+
+fun mockFirebaseDynamicLinks(): FirebaseDynamicLinks {
+    val dynamicLinks = Mockito.mock(FirebaseDynamicLinks::class.java)
+    `when`(dynamicLinks.getDynamicLink(any<Intent>())).then {
```
ValentinAebi, line 14: No need to change this, but for future work if you don't know it already, I realized yesterday that there is a `whenever` method that does the same as `when` but is more handy to use as no ` are required.

app/src/test/java/com/github/multimatum_team/multimatum/util/MockFirebaseDynamicLinks.kt
```diff
@@ -0,0 +1,33 @@
...
+        val intent = it.getArgument<Intent>(0)
+        var dynamicLink: String? = null
+        var deepLink: String? = null
+        if (intent.data != null) {
+            dynamicLink = intent.data.toString()
+            deepLink = intent.data!!.getQueryParameter("link")!!
+        }
```
ValentinAebi, line 21:
```suggestion
        val dynamicLink: String? = intent.data?.toString()
        val deepLink: String? = intent.data?.getQueryParameter("link")!!
```

app/src/test/java/com/github/multimatum_team/multimatum/util/MockFirebaseDynamicLinks.kt
felko: Thanks for the tip

app/src/main/java/com/github/multimatum_team/multimatum/activity/GroupInviteActivity.kt
felko: Sure, I extracted the code and added comments to explain the logic

ValentinAebi (approved):

PR #259, opened 2022-05-17
Customise Calendar View

### LINKED ISSUE #250: Show the deadline in the calendar view

Matathiasq13824:
This PR add in the `CalendarView` dots below date to see which one has already a deadline or not.
This was more difficult than expected because of the very few library that implement this and can be used using Kotlin.
There is still a problem that I haven't figured out how to solve in the dark mode of the calendar view

This close #250

### DIFF
--- a/app/src/main/java/com/github/multimatum_team/multimatum/activity/CalendarActivity.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/activity/CalendarActivity.kt
@@ -2,7 +2,6 @@ package com.github.multimatum_team.multimatum.activity
 
 import android.graphics.Color
 import android.os.Bundle
-import android.util.Log
 import android.view.MotionEvent
 import android.widget.Button
 import androidx.activity.viewModels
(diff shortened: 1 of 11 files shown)


macintosh1398 (approved):
Very good job on this PR. It is great because you managed to add a feature that I did not succeed to add. It is a simple view, but just to compare, the IOS view is basically the same. The code is clear, modular and well commented. Speaking about the comments, I would be good to check the upper cases of the comment and prefer the /*blabla*/ syntax when you have multiple lines.

app/src/main/java/com/github/multimatum_team/multimatum/activity/CalendarActivity.kt
```diff
@@ -1,58 +1,105 @@
...
+            // with the 3 override functions
+            .setCallback(this)
+
+        updateColorText()
+        setDotForEveryDeadline()
         initTextInput()
         initAddButton()
     }
 
+    private fun updateColorText(){
```
felko, line 71: Small formatting issue, please put a space between `)` and `{`.
Sorry I coudln't come up with something more interesting to say, code looks good otherwise

felko (approved):
Great job, thanks for fixing a bunch of warnings.
My only comment isn't really an issue so I'm already approving.

PR #260, opened 2022-05-18
Adding Firebase storage and display of location

macintosh1398:
This PR give the user the ability to select a location for a particular deadline and the location name and coordinates (GeoPoint) will be stored on Firebase. When the user looks at the deadline details, the location name will appear (if there is one). Some UI improvements/icon addition were also made (main screen and deadline details).

### DIFF
--- a/app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
@@ -322,7 +322,6 @@ class DeadlineDetailsActivity : AppCompatActivity() {
                 getString(R.string.DueTheXatX, dateTime.toLocalDate(), dateTime.toLocalTime())
             titleView.text = SpannableStringBuilder(title)
             if (deadline.locationName != null) locationView.text = deadline.locationName
-            locationView.text = deadline.locationName
             descriptionView.text = SpannableStringBuilder(description)
             doneButton.isChecked = (state == DeadlineState.DONE)
             updateDetail()
--- a/app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
@@ -10,6 +10,7 @@ import android.content.SharedPreferences
 import android.graphics.Color
 import android.os.Bundle
 import android.text.SpannableStringBuilder
+import android.util.Log
 import android.view.View
 import android.widget.*
 import androidx.activity.viewModels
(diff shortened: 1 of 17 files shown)


macintosh1398:
New test was added for serialization to have full coverage. But in the other hand, I had to remove AddDeadlineActivity tests in order to solve the testing bug of mapbox. This results in a big drop of code coverage. Needs to be solved.

app/src/main/java/com/github/multimatum_team/multimatum/activity/MainActivity.kt
```diff
@@ -23,6 +23,7 @@ import com.github.multimatum_team.multimatum.repository.PdfRepository
 import com.github.multimatum_team.multimatum.util.DeadlineNotification
 import com.github.multimatum_team.multimatum.viewmodel.DeadlineListViewModel
 import com.github.multimatum_team.multimatum.viewmodel.GroupViewModel
+import com.google.firebase.firestore.FirebaseFirestore
```
Matathiasq13824, line 26: Does this import is still useful ?

Matathiasq13824 (approved):
Great job by finding a way to implement this without removing the AddDeadline tests. The coverage is still low but it can be solved next sprint. I have find only a little problem that I have commented but It is too small to not approve now

app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
```diff
@@ -318,6 +321,8 @@ class DeadlineDetailsActivity : AppCompatActivity() {
             dateView.text =
                 getString(R.string.DueTheXatX, dateTime.toLocalDate(), dateTime.toLocalTime())
             titleView.text = SpannableStringBuilder(title)
+            if (deadline.locationName != null) locationView.text = deadline.locationName
+            locationView.text = deadline.locationName
```
ValentinAebi, line 325: I think there is a problem here...

ValentinAebi (changes requested):
Congrats on fixing the bug! 🥇 
Which solution worked in the end?

app/src/main/java/com/github/multimatum_team/multimatum/activity/MainActivity.kt
macintosh1398: No, removed

app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
macintosh1398: Oh yes your're right. Thanks. About Mapbox, I created a new activity called SearchLocation that contains the mapbox view. This activity is lauched using the button on the addDeadlineActivity and the location result is communicated via intent. This is done using the StartActivityForResult() function.

macintosh1398: So it this "solution" give us the ability to execute addDeadlineTests without any mapbox stuff in it.

ValentinAebi: OK nice. I am just wondering where the coverage drop comes from... From CodeClimate I see that `SearchLocationActivity` has 0% coverage, is that correct?

ValentinAebi (approved):

PR #263, opened 2022-05-19
Download and open pdf

LennyDZ:
add in deadline detail the possibility to download and open the pdf associated with a deadline

Not covered by any tests yet.

### DIFF
--- a/app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
@@ -326,9 +326,10 @@ class DeadlineDetailsActivity : AppCompatActivity() {
             val description = deadline.description
             val group = deadline.owner
 
+            LogUtil.debugLog(pdfLink.toString())
             //display download button if there's a pdf
             if (pdfLink != "") {
-                downloadLinkText.text = PDFUtil.getFileNameFromUri(pdfLink.toUri(), this)
+                downloadLinkText.text = PDFUtil.getFileNameFromUri(pdfLink.toUri(), this).drop(16)
             } else {
                 downloadLinkText.visibility = View.INVISIBLE
             }
@@ -461,7 +462,7 @@ class DeadlineDetailsActivity : AppCompatActivity() {
     fun downloadPdf(view: View) {
         pdfRepository.downloadPdf(
             pdfLink,
-            PDFUtil.addRdmCharToStr(downloadLinkText.text.toString())
+            downloadLinkText.text.toString()
         ) {
             LogUtil.debugLog(it.path)
             val intent = Intent(Intent.ACTION_VIEW)
--- a/app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
@@ -8,7 +8,6 @@ import android.content.Intent.EXTRA_TEXT
 import android.graphics.Color
 import android.os.Bundle
 import android.text.SpannableStringBuilder
-import android.util.Log
 import android.view.View
 import android.widget.*
 import androidx.activity.viewModels
@@ -23,7 +22,6 @@ import com.github.multimatum_team.multimatum.model.DeadlineState
 import com.github.multimatum_team.multimatum.model.GroupOwned
 import com.github.multimatum_team.multimatum.model.UserOwned
 import com.github.multimatum_team.multimatum.repository.DeadlineID
-import com.github.multimatum_team.multimatum.repository.FirebasePdfRepository
 import com.github.multimatum_team.multimatum.repository.PdfRepository
 import com.github.multimatum_team.multimatum.service.ClockService
 import com.github.multimatum_team.multimatum.util.DeadlineNotification
@@ -31,9 +29,6 @@ import com.github.multimatum_team.multimatum.util.JsonDeadlineConverter
 import com.github.multimatum_team.multimatum.util.PDFUtil
 import com.github.multimatum_team.multimatum.viewmodel.DeadlineListViewModel
 import com.github.multimatum_team.multimatum.viewmodel.GroupViewModel
-import com.google.firebase.ktx.Firebase
-import com.google.firebase.ktx.initialize
-import com.google.firebase.storage.FirebaseStorage
 import dagger.hilt.android.AndroidEntryPoint
 import java.time.Duration
 import java.time.LocalDateTime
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
```diff
@@ -43,6 +51,9 @@ class DeadlineDetailsActivity : AppCompatActivity() {
     @Inject
     lateinit var clockService: ClockService
 
+    @Inject
+    lateinit var firebasePdfRepository: PdfRepository
```
felko, line 55: Nitpick but the pdf repository isn't necessarily linked to firebase, since we mock it in tests.
The fact that it is a firebase implementation should be irrelevant to the activity.
It would also be more consistent with the rest of the code.

app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
```diff
@@ -15,24 +13,34 @@ import android.view.View
...
+import com.github.multimatum_team.multimatum.repository.PdfRepository
 import com.github.multimatum_team.multimatum.service.ClockService
 import com.github.multimatum_team.multimatum.util.DeadlineNotification
 import com.github.multimatum_team.multimatum.util.JsonDeadlineConverter
+import com.github.multimatum_team.multimatum.util.PDFUtil
 import com.github.multimatum_team.multimatum.viewmodel.DeadlineListViewModel
 import com.github.multimatum_team.multimatum.viewmodel.GroupViewModel
+import com.google.firebase.ktx.Firebase
+import com.google.firebase.ktx.initialize
+import com.google.firebase.storage.FirebaseStorage
```
felko, line 36: Android code should not depend on firebase at all.

app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
```diff
@@ -57,20 +68,26 @@ class DeadlineDetailsActivity : AppCompatActivity() {
...
     private var dateTime: LocalDateTime = LocalDateTime.of(2022, 10, 10, 10, 10)
     private var state: DeadlineState = DeadlineState.TODO
+    private var pdfLink: String = ""
 
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_deadline_details)
 
+        Firebase.initialize(this)
```
felko, line 85: The app was already initialized in `MainActivity`

app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
```diff
@@ -57,20 +68,26 @@ class DeadlineDetailsActivity : AppCompatActivity() {
...
+        firebasePdfRepository = FirebasePdfRepository(FirebaseStorage.getInstance())
```
felko, line 86: Use dependency injection instead of manually instantiating the repository, otherwise we cannot test it

app/src/main/java/com/github/multimatum_team/multimatum/activity/DeadlineDetailsActivity.kt
```diff
@@ -301,10 +318,16 @@ class DeadlineDetailsActivity : AppCompatActivity() {
             val deadline = deadlines[id]!!
             dateTime = deadline.dateTime
             state = deadline.state
+            pdfLink = deadline.pdfPath
             val title = deadline.title
             val description = deadline.description
             val group = deadline.owner
 
+            //display download button if there's a pdf
+            if (pdfLink != "") {
```
felko, line 327: Please do not represent absent values by empty strings but by null (using `String?`).

app/src/main/java/com/github/multimatum_team/multimatum/repository/FirebasePdfRepository.kt
```diff
@@ -20,9 +22,10 @@ class FirebasePdfRepository @Inject constructor(database: FirebaseStorage) : Pdf
         if (data != Uri.EMPTY) {
             val ref =
                 storageReference.child(
-                    FirebaseAuth.getInstance().uid + "/upload" + (Int.MIN_VALUE..Int.MAX_VALUE).random()
-                        .toString() + PDFUtil.getFileNameFromUri(
-                        data
+                    FirebaseAuth.getInstance().uid + "/" + PDFUtil.addRdmCharToStr(
```
felko, line 25: Please do `@Inject constructor(auth: FirebaseAuth, database: FirebaseStorage)` instead of getting the instance, this way we can test it by replacing the auth by a mocked object

app/src/main/java/com/github/multimatum_team/multimatum/util/PDFUtil.kt
```diff
@@ -25,9 +22,28 @@ object PDFUtil {
...
+                cursor.getString(cursor.getColumnIndexOrThrow(OpenableColumns.DISPLAY_NAME))
+            cursor.close()
+        } else {
+            val pathStr = pdfData.toString()
+            fileName = pathStr.substring(pathStr.lastIndexOf('/') + 1).drop(16)
+        }
+        return fileName
+    }
+
+    fun addRdmCharToStr(input: String, numOfChar: Int): String {
```
felko, line 44: This is not a good way to generate unique identifiers as it can generate the same ID twice (even if unlikely).
Generating unique IDs is not simple to do, you could rely on the `UUID` class in the java standard library.
<https://developer.android.com/reference/java/util/UUID>

felko (changes requested):
Good work. By applying the requested changes, you should be able to test your code more easily and increase coverage.
There seems to be a bug when trying to download a linked PDF, I can laucnh my PDF viewer (I'm using Foxit) but the file is not opened and I can't seem to find it.
A possible solution to generate a unique file would be to use temporary storage (see <https://stackoverflow.com/questions/3425906/creating-temporary-files-in-android>).
As a last remark, I noticed that when no PDF files are attached to a deadline, the button shows "Download linked pdf" despite there being no pdf, it would be nice to change that message in that case.

app/src/main/java/com/github/multimatum_team/multimatum/activity/MainActivity.kt
```diff
@@ -53,7 +54,6 @@ class MainActivity : AppCompatActivity() {
         super.onCreate(savedInstanceState)
 
         Firebase.initialize(this)
-        pdfRepository = FirebasePdfRepository(FirebaseStorage.getInstance())
```
felko, line 56: Please use dependency injection
```kt
@Inject
lateinit var pdfRepository: PdfRepository
```
with the repository provider in `DependencyProvider.kt`

felko (changes requested):

macintosh1398 (approved):

PR #268, opened 2022-05-19
Fix bug in `DeadlineDetailsActivity` causing the app to crash when clicking on a group owned deadline

### LINKED ISSUE #266: Fix bug in `DeadlineDetailsActivity` causing the app to crash when trying to display a group owned deadline
I think there was a merge problem that duplicated the code that displays the deadline owner in the `DeadlineDetailsActivity`, that causes the activity to call the group viewmodel outside a group viewmodel observer, which makes the app crash since the group viewmodel doesn't have time to fetch the groups before trying to get the group name.

felko:
See linked issue for a more detailed description of the bug.

Closes #266


ValentinAebi (approved):

Voxanimus (approved):
I'm not sure to understand why these lines are the cause of the bug. If you can explain it to me it's cool but anyway I trust you.

felko:
> I'm not sure to understand why these lines are the cause of the bug. If you can explain it to me it's cool but anyway I trust you.

I explained in the issue

PR #269, opened 2022-05-19
Feature/create group

Voxanimus:
For some obscur reason the test aren't passing but everything is like other tests

### DIFF
--- a/app/src/main/java/com/github/multimatum_team/multimatum/activity/GroupsActivity.kt
+++ b/app/src/main/java/com/github/multimatum_team/multimatum/activity/GroupsActivity.kt
@@ -1,20 +1,20 @@
 package com.github.multimatum_team.multimatum.activity
 
 import android.content.DialogInterface
-import android.content.Intent
 import android.os.Bundle
 import android.view.View
 import android.widget.EditText
 import android.widget.LinearLayout
 import android.widget.ListView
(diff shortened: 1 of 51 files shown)


app/src/main/java/com/github/multimatum_team/multimatum/activity/GroupsActivity.kt
```diff
@@ -41,4 +47,39 @@ class GroupsActivity : AppCompatActivity() {
...
+        alertDialogBuilder.setView(input)
+
+        //add cancel and create button
+        alertDialogBuilder.setPositiveButton("Create") { dialog, _ ->
+            createGroup(dialog, input)
+        }
+        alertDialogBuilder.setNegativeButton("Cancel") { dialog, _ ->
+            dialog.dismiss()
+        }
+        alertDialogBuilder.create()
```
Matathiasq13824, line 72: This function is not necessary as the show() call this function too but this a nitpick as it doesn't change anything

Matathiasq13824 (approved):
It is quite sad that the test doesn't pass but I have checked and it doesn't make sense, it should work. We need to correct this next week

felko (approved):
Good work, nothing to add over Matathiasq13824's review.
Try to write test for next sprint

PR #274, opened 2022-05-23
Tests/groups activity

### LINKED ISSUE #272: Tests for `GroupsActivity`

ValentinAebi:
A new test for `GroupsActivity` that checks how `addGroup` configures the `AlertDialog` and the created group. This test uses mocking (it is not an end-to-end test) and is designed as a regression test rather than a way to test the current setup. I created a testcase `GroupsActivityTest2`, distinct from `GroupsActivityTest`, because these two test cases must inject different dependencies/mocks.

Close #272


LennyDZ (approved):
Seems all good for me

macintosh1398 (approved):
Good work! Very relevant tests, and good helper functions. Everything is clear, you got my approval.

PR #290, opened 2022-05-25
Test/display location activity

### LINKED ISSUE #289: Tests for `DisplayLocationActivity`

ValentinAebi:
Test for `DisplayLocationActivity`. This test (ab)uses DI, injecting functions that take other functions as an argument... The purpose of this confusing pattern is to disable some function calls inside `onCreate` when the function is run in tests (because they throw exceptions otherwise). This is a bit ugly, but it is the only way that I found to make this test work, and I think that despite that the test is meaningful.

Close #289

### DIFF
--- a/app/src/test/java/com/github/multimatum_team/multimatum/DisplayLocationActivityTest.kt
+++ b/app/src/test/java/com/github/multimatum_team/multimatum/DisplayLocationActivityTest.kt
@@ -1,0 +1,123 @@
+package com.github.multimatum_team.multimatum
+
+import android.content.Context
+import android.content.Intent
+import androidx.test.core.app.ActivityScenario
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.espresso.intent.Intents
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.github.multimatum_team.multimatum.activity.DisplayLocationActivity
+import com.mapbox.maps.CameraOptions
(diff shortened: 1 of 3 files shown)


Voxanimus (commented):
Are you sure that one test is sufficient to test this feature ?

ValentinAebi:
> Are you sure that one test is sufficient to test this feature ?

Of course it would be better to have several tests, but this one already gives a good coverage on the class. And it would be difficult to split it into several tests.

Voxanimus (approved):

app/src/test/java/com/github/multimatum_team/multimatum/DisplayLocationActivityTest.kt
```diff
@@ -0,0 +1,123 @@
...
+    object TestDependencyProvider {
+
+        @Singleton
+        @Provides
+        fun provideMapViewProducer(): MapViewProducer =
+            MapViewProducer { mockMapView }
+
+        @Singleton
+        @Provides
+        fun provideContentViewSetter(): ViewActionPerformer =
```
felko, line 117: Are the two provides logically coupled? When I put every repository in the FirebaseRepositoryModule, it turned out to be inconvenient in tests since I had to provide all repositories in the test module even when they were not used. @LennyDZ and I had to update almost all the tests everytime we added a new type of repository.

felko (approved):

PR #295, opened 2022-05-25
add offline behavior for pdf

LennyDZ:
I add some check before upload and download on wether the app is connected to internet. If not, the deadline will be added without the pdf for upload, and pdf will not be openable in deadlineDetail. In both case, an alert dialog pop to warn the user.


app/src/androidTest/java/com/github/multimatum_team/multimatum/util/MockPdfRepository.kt
```diff
@@ -8,15 +8,15 @@ import org.mockito.kotlin.doNothing
...
         doNothing()
     }
 
     override fun delete(path: String) {
         doNothing()
     }
 
-    override fun downloadPdf(path: String, title: String, callback: (File) -> Unit) {
-        callback(any())
+    override fun downloadPdf(path: String, title: String, callback: (File?, Boolean) -> Unit) {
```
felko, line 19: Same here

felko (approved):

Matathiasq13824 (approved):
Nothing to add with what felko has said

PR #300, opened 2022-05-26
Prevent anonymous users from accessing the groups page

felko:
To prevent anonymous users from creating groups which makes the app crash, the "groups" button on the main page now checks whether the user is signed-in or anonymous. In the former case, it behaves as before, starting the groups activity, but in the latter, an error message is displayed to signal the user that the groups feature is reserved to signed-in users. The user can either dismiss the message or click on a button to be redirected to the sign-in page.


Voxanimus (approved):
Nothing to say. It's a very simple way to implement the review and test are good so well done.

LennyDZ (approved):

PR #301, opened 2022-05-26
Ignore deserialization errors from Firebase

### LINKED ISSUE #299: Ignore deserialization errors from Firebase
When deserializing data fetched from Firebase, our code currently performs unchecked casts and other unsafe operations, crashing when the data is incorrect.
A more sensible way of handling things would be to simply ignore incorrectly formatted user information, deadlines and groups, so the app doesn't crash when it happens.
However, the fact that it currently crashes turns out to be a useful tool to find errors in our code, so we should still emit a warning when it happens.

felko:
As described in #299, we should avoid that ill-formed data makes the application crash, especially since there is no way for the user to clear the Firebase cache. This PR prevents the app from crashing simply by filtering out deadlines and groups that couldn't be parser.
The code in `FirebaseDeadlineRepository.deserializedDeadline` isn't very good, but I couldn't figure out how to write it in a better way, I'm open to suggestions.

Closes #299

### DIFF
--- a/app/src/androidTest/java/com/github/multimatum_team/multimatum/util/MockGroupRepository.kt
+++ b/app/src/androidTest/java/com/github/multimatum_team/multimatum/util/MockGroupRepository.kt
@@ -40,8 +40,8 @@ class MockGroupRepository(initialContents: List<UserGroup>) : GroupRepository() 
             })
         }
 
-    override suspend fun fetch(id: GroupID): UserGroup =
-        groups[id]!!
+    override suspend fun fetch(id: GroupID): UserGroup? =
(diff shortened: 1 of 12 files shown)


Matathiasq13824 (commented):
I see a lot of warning about a lot of return, is it possible to remove them ?

felko:
> I see a lot of warning about a lot of return, is it possible to remove them ?

I don't really see a better way, as mentionned in the PR I am open for suggestions but otherwise we can try to figure it out for next sprint.

LennyDZ (approved):

PR #302, opened 2022-05-26
Fix/crash when log out without conexion

Voxanimus:
There was crashed when trying to log out without internet connection. At begining I tried to make complicated solution by letting user log out and so getting logged with a "false" anonymous user. It leads to complicated stuff with checking connection in while loop asynchronously so I choose a way more simple solution. When the user tried to log out without internet solution he cannot log out and a toast is displayed to inform him.


app/src/main/java/com/github/multimatum_team/multimatum/activity/MainActivity.kt
```diff
@@ -20,6 +20,7 @@ import com.github.multimatum_team.multimatum.adaptater.NoFilter
 import com.github.multimatum_team.multimatum.repository.DeadlineRepository
 import com.github.multimatum_team.multimatum.repository.PdfRepository
 import com.github.multimatum_team.multimatum.util.DeadlineNotification
+import com.github.multimatum_team.multimatum.viewmodel.AuthViewModel
```
Matathiasq13824, line 23: Is this necessary ?

app/src/main/java/com/github/multimatum_team/multimatum/activity/ProfileActivity.kt
```diff
@@ -34,8 +38,12 @@ class ProfileActivity : AppCompatActivity() {
 
         logOutButton.setOnClickListener {
             Log.d(TAG, "Logging out...")
-            userViewModel.signOut()
-            AuthUI.getInstance().signOut(this)
+            if((this.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager).activeNetwork != null) {
```
Matathiasq13824, line 41: Does this could be tested or you don't have the time ?

app/src/main/java/com/github/multimatum_team/multimatum/viewmodel/AuthViewModel.kt
```diff
@@ -1,5 +1,6 @@
 package com.github.multimatum_team.multimatum.viewmodel
 
+import android.widget.Toast
```
Matathiasq13824, line 3: Does this is necessary ?

Matathiasq13824 (changes requested):
Good job, I just hae some question but everything is good otherwise

app/src/main/java/com/github/multimatum_team/multimatum/activity/ProfileActivity.kt
```diff
@@ -34,8 +38,12 @@ class ProfileActivity : AppCompatActivity() {
...
+                userViewModel.signOut()
+                AuthUI.getInstance().signOut(this)
+            }else{
+                Toast.makeText(this, "You cannot log out without internet connection", Toast.LENGTH_SHORT).show()
```
LennyDZ, line 45: Maybe make an alertDialog instead of a toast, it'ld make more sense I think

LennyDZ (commented):
Just make sure to check unused import as pointed.

app/src/main/java/com/github/multimatum_team/multimatum/activity/ProfileActivity.kt
Voxanimus: I think not because with alert dialog you have 2 button that doesn't have any purpose in this case so a toast seems to be a good solution in this case.

app/src/main/java/com/github/multimatum_team/multimatum/activity/ProfileActivity.kt
Voxanimus: It depends on weird case where we don't have internet connection and I'm not sure that we could "disable" internet connection in test and If it's possible search how it's works will take me too much time

Matathiasq13824 (approved):

LennyDZ (approved):

